Issue #5, opened 2021-11-19
How to override default mappings


j-hui:
I'm trying to override the default mappings given by this extension. For example, I want to remape `<C-f>` to `<Right>`. However, I am unable to do so.

This is an extract from my config (I have a lot of other mappings):

```vim
    local fb_actions = require 'telescope'.extensions.file_browser.actions
    local imaps = {
      ['<C-f>'] = { '<Right>',  type = 'command' },
    }
    require'telescope'.setup {
      extensions = {
        file_browser = { mappings = { i = imaps, }, },
      },
    }
```

I was looking through the source code for this plugin, and saw that the fb_actions were being mapped in [`attach_mappings`](https://github.com/nvim-telescope/telescope-file-browser.nvim/blob/master/lua/telescope/_extensions/file_browser/picker.lua#L92). This runs after extension `mappings` are set, meaning my mappings are being overriden by this plugin.

I'm still new to Telescope and the way things are organized, so apologies if this is already answered elsewhere, but how can I override `attach_mappings` this to disable these mappings? And is there any way I can do this without having to copy and paste in the `replace_if` chunk of code? (I'm not actually sure what that's doing.)

fdschmidt93:
Thanks for the issue!

Your `setup` should work, but in trying to debug this I've realized some oddities.

Are you using something like https://github.com/tjdevries/astronauta.nvim to do keymappings in lua and pass an actual lua function? What may happen (depending on your startup config), that the file browser is mapped before the extension is set up and therefore it gets mapped without config.

I've had that issue and now map it via a viml command and then overriding keymaps as intended. 

Furthermore, it looks like unmapping does not seem to work though?

`["<C-f>"] = false` should unmap `<C-f>`. 

I'll look into this more.

fdschmidt93:
I think being unable to unmap mappings that are attached by default like so

https://github.com/nvim-telescope/telescope-file-browser.nvim/blob/a9d361cac3194ba66e76ddf96ee9b2c7e7ab9a73/lua/telescope/_extensions/file_browser/picker.lua#L79-L111

is an upstream bug. 

Could you please confirm that the below holds for you? Specifically, mapping `<C-a>` to `false` doesn't achieve anything while mapping it to the `print` function overrides the mapping as intended?

```lua
require "telescope".setup {
  pickers = {
    git_branches = {
      mappings = {
        i = {
          ["<C-a>"] = false, -- should unmap create branch, does not
[...]

PR #6, opened 2021-11-19
fix: keymapping defaults and overrides

fdschmidt93:
* Consistent default keymappings between insert and normal mode
* Fix being unable to fully override (unmap) default keymappings

This partially solves #5 as now mappings can be fully unmapped and overriden at telescope setup. What's still not possible is unmapping when called interactively eg

```lua
require "telescope".extensions.file_browser.file_browser {
  attach_mappings = function(prompt_bufnr, map)
    map("i", "<C-f>", false)
    return true
  end
}
```
which seems to be an upstream issue.

In any case, I prefer having the default configuration in `pconf` and it should largely fix the issue.

@j-hui could you please test whether this works for you? Many thanks in advance!

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,11 +75,15 @@ vim.api.nvim_set_keymap(
 | `<C-y>/y`        | Copy (multi-selected) files or folders to cwd        |
 | `<C-d>/dd`       | Delete (multi-selected) files or folders             |
 | `<C-r>/r`        | Rename (multi-selected) files                        |
+| `<C-e>/e`        | Add File/Folder at cwd; trailing `/` creates folder  |
 | `--/m`           | Move multi-selected files to cwd                     |
 | `<C-h>/h`        | Toggle hidden files                                  |
 | `<C-o>/o`        | Open file with default system application            |
+| `<C-g>/g`        | Go to parent directory                               |
 
 Copying and moving files first requires you to multi-select your files and folders and then moving to the target directory to copy and move the selections to (cf. [demo](#demo)). Renaming multi-selected files or folders launches batch renaming, which enables to user to rename or move multiple files at once, as moving files is analogous to renaming the file.
+
+As a tip: you can use telescope's `which_key` action mapped by default to `<C-/>` and `?` in insert and normal mode, respectively, to inspect the mappings attached to the picker from within telescope.
 
 ## Documentation
 
--- a/doc/telescope-file-browser.txt
+++ b/doc/telescope-file-browser.txt
@@ -32,8 +32,7 @@ require "telescope".extensions.file_browser
 <
 In particular:
 - `file_browser`: constitutes the main picker of the extension
-- `picker`: unconfigured equivalent of `file_browser` defaulting to extensiond
-  defaults
+- `picker`: unconfigured equivalent of `file_browser`
 - `actions`: extension actions make accessible for remapping and custom usage
 - `finder`: low-level finders -- if you need to access them you know what you
   are doing
@@ -54,7 +53,7 @@ fb_picker.file_browser({opts})                      *fb_picker.file_browser()*
     Default keymaps in insert/normal mode:
       - `<cr>`: opens the currently selected file, or navigates to the
         currently selected directory
-      - `<C-e>`: creates new file in current directory, creates new directory
+      - `<C-e/e>`: creates new file in current directory, creates new directory
         if the name contains a trailing '/'
         - Note: you can create files nested into several directories with
           `<C-e>`, i.e. `lua/telescope/init.lua` would create the file
@@ -68,7 +67,7 @@ fb_picker.file_browser({opts})                      *fb_picker.file_browser()*
         current directory
       - `<C-f>/f`: toggle between file and folder browser
       - `<C-h>/h`: toggle hidden files
-      - `<C-d/dd>`: remove currently or multi selected file(s) or folder(s)
+      - `<C-d>/dd`: remove currently or multi selected file(s) or folder(s)
         recursively
       - --/m`: move multi selected file(s) or folder(s) recursively to current
         directory in file browser
@@ -193,8 +192,8 @@ fb_actions.open_file()                                *fb_actions.open_file()*
 
 
 
-fb_actions.goto_prev_dir({prompt_bufnr}, {bypass}) *fb_actions.goto_prev_dir()*
-    Goto previous directory in |builtin.file_browser|.
+fb_actions.goto_parent_dir({prompt_bufnr}, {bypass}) *fb_actions.goto_parent_dir()*
+    Goto parent directory in |builtin.file_browser|.
 
 
     Parameters: ~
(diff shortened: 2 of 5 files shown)


j-hui:
@fdschmidt93 I'm on this branch and it doesn't seem to be working for me. I'm currently testing with:

```lua
    require'telescope'.setup {
      extensions = {
        file_browser = {
          mappings = {
            i = {
              ['<C-f>'] = function () print("C-f overriden") end,
            }
          },
        },
      },
    }
```

Is there something I should be doing differently to remap `<C-f>`?

fdschmidt93:
I checked your snippet and it works as intended for me.

> My vimrc setup is a bit of an over-engineered mess with callbacks out the wazoo, but all the telescope-specific mappings are being set using the mappings mechanism that it provides. (Though I'm not really sure if that's the right thing to do.)

I think this is a "access" order issue. In brief, order should be

setup (the above) -> `require("telescope").load_extension "file_browser"` *actually loads the config) -> everything else

Before I go on a wild goose chase, this is setup as intended? :sweat_smile:

Another way to validate would be: keep just the setup and load, unmap the file browser and try this from cmdline

`require "telescope".extensions.file_browser.file_browser()`

----

On this PR itself, I'll have to check https://github.com/nvim-telescope/telescope.nvim/pull/1473 which cleans up mappings a lot upstream

ikerurda:
Hi!
I'm not able to overwrite mappings either.
Currently I have a similar config to the following:
```lua
local tl = require "telescope"
local fb_actions = require"telescope".extensions.file_browser.actions
local actions = require"telescope.actions"

tl.setup {
  extensions = {
    file_browser = {
      theme = "ivy",
      layout_config = {height = .5},
      mappings = {
        i = {["<C-v>"] = fb_actions.move_file, ["<C-x>"] = fb_actions.remove_file, ["<C-d>"] = actions.preview_scrolling_down}
      }
    }
  }
}
tl.load_extension "file_browser"
```
Even when trying it from the commandline, as you suggested, I only get the plugin with that `theme` and `layout_config` when `mappings = {...}` and `local fb_actions = ...` are removed

fdschmidt93:
Ah, I've just realized what the issue most likely is.

You guys do `local fb_actions = require"telescope".extensions.file_browser.actions` and thereby access the extension de facto before it has been set up. Then, the picker config table never gets properly created and forwarded onto the picker from `require "telescope".extensions.file_browser.file_browser` because the setup is called on the first access onto the extension.

I think this is something we might have to fix upstream. Right now this can be fixed by wrapping the action into another function, such that `fb_actions` will be required lazily or via hot-reloading through plenary.

```lua

if pcall(require, "plenary") then
  R = function(name)
    require("plenary.reload").reload_module(name)
    return require(name)
  end
end
```

and then you'd map

`["key"] = R "telescope".extensions.file_browser.actions`

I'll try to look into it some time next week.

j-hui:
> setup (the above) -> require("telescope").load_extension "file_browser" *actually loads the config) -> everything else

Yeah, this is the verbatim lua snippet I use to set things up:

```lua
    local fb_actions = require 'telescope'.extensions.file_browser.actions
    local imaps = {
      ['<C-a>'] = { '<Home>',   type = 'command' },
      ['<C-e>'] = { '<End>',    type = 'command' },
      ['<C-f>'] = { '<Right>',  type = 'command' },
      ['<C-b>'] = { '<Left>',   type = 'command' },
      ['<C-h>'] = { '<BS>',     type = 'command' },
      ['<C-k>'] = function () vim.cmd 'norm! d$' end,
      ['<Esc>'] = require'telescope.actions'.close,
      ['<C-c>'] = { '<Esc>',    type = 'command' },
      ["<C-e>"] = function() print("test") end, -- successfully overrides
    }
    require'telescope'.setup {
      defaults = { mappings = { i = imaps } },
      extensions = {
        file_browser = {
          mappings = {
            i = {
              ['<C-f>'] = function () print("C-f overriden") end, -- doesn't work
            }
            -- vim.tbl_extend('force', imaps, {
            --   -- ['<C-j>'] = fb_actions.create_file,
            --   ['<C-j>'] = function () print("c-j") end,
            --   ['<C-v>'] = fb_actions.toggle_browser,
            -- })
          },
        },
      },
    }
  require'telescope'.load_extension 'file_browser'
```

I'm not sure if tihs is related to requiring `fb_actions` early though, because I'm not even able to override `<C-f>` with a regular function that does not depend on `fb_actions`.

This is also true when I check out `apply_mappings_fixup`.

fdschmidt93:
Yes, it's because require fb_actions is on the top.

> because I'm not even able to override <C-f> with a regular function that does not depend on fb_actions.

That doesnt do anything because you index the extensions table via requiring fb_actions.

You need to wrap every reference to in a function like

`[key] = function(prompt_bufnr) fb_action.remove_file(prompt_bufnr) end`

or hot reloading as mentioned above.

Sorry, typing from mobile.

j-hui:
ahhh I see what you mean. This works:

```lua
    local function fb_action(f)
      return function (b)
        require'telescope'.extensions.file_browser.actions[f](b)
      end
    end
    local imaps = {
      -- same as before
    }
    require'telescope'.setup {
      defaults = { mappings = { i = imaps } },
      extensions = {
        file_browser = {
          mappings = {
            i = vim.tbl_extend('force', imaps, {
              ['<C-j>'] = fb_action 'create_file',
              ['<C-v>'] = fb_action 'toggle_browser',
            }),
          },
        },
      },
    }
    require'telescope'.load_extension 'file_browser'
```

fdschmidt93:
> The definition of fb_action is a little clunky but it keeps the mappings section cleaner at least

Yeah, I've realized we need to revamp extension setup a bit upstream to make this use case a bit more simple. I think we can also get rid of `require "telescope".load_extension "file_browser"` if we do it properly.

> Still learning Lua, but it seems like require'pkg' can be side-effectful? And if you evaluate that expression multiple times, will it initialize things multiple times?

AFAIK, lua caches `require`. This shouldn't be so much of an issue here, it's just that the telescope extension framework is written in a way that the extension setup is only called once on the first access of the `require "telescope".extension.$EXTENSION` table. I guess you could call afterwards with a hidden API once more, but this is not user friendly.

In any case, telescope's plan is to move pickers out of core incrementally (and the file-browser being a primary case), we need to polish that extension system.

Apologies, you'll have to live with the work-around for a bit, while fixing the upstream in a simple PR is easy, the general API needs a bit of a cleaner redesign.

j-hui:
Got it, thanks! I can just stick with my current workaround, and am happy to close this issue for now.

> Don't think you have to wrap it in tbl_extend force, no?

I'm not sure if this is intentional upstream, but I first started doing this because I noticed that my global mappings were being overridden by extension-specific mappings when I was trying out the built-in file browser. That is, default globals < user globals < default extension < user extension. (Is that correct?)

My tbl_extend force makes it such that my own mappings are also being declared as extension-specific mappings, so that they take the highest priority.

fdschmidt93:
I think for the time being, this PR is a strict improvement wrt the mappings situation. Maybe if the upstream mappings PR is merged, we'll have to change something, but that may or may not take some more time.

Hopefully no breakages for users, my testing (and `actions.which_key`) indicate that by default everything works as intended.

Slight breaking change: `fb_actions.goto_prev_dir` has now been renamed to `fb_actions.goto_parent_dir`

PR #11, opened 2021-11-30
Add mapping to table of add file/folder plus note about default mapping

benfrain:
Updated the table to include the default mapping for adding a file/folder.
Also tweaked the text a litte to point out there is no default set for Move.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,15 +69,16 @@ vim.api.nvim_set_keymap(
 
 ## Mappings
 
-| Insert / Normal  | Action                                               |
-|------------------|------------------------------------------------------|
-| `<C-f>/f`        | Toggle between file and folder browser               |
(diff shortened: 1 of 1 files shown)


fdschmidt93:
Many thanks :)

I'm afraid it conflicts a bit with https://github.com/nvim-telescope/telescope-file-browser.nvim/pull/6/files where I already address some of these things -- so I plan on incorporating this PR one way or another soon enough.

At the same time I need to properly test https://github.com/nvim-telescope/telescope.nvim/pull/1473 before I want to settle on mappings (and how they are set by default here in `telescope-file-browser`).

I'll most likely be will be keeping telescope involvement at a bare minimum practically speaking up until end of February (full force on work til end of this year + travelling almost entire January, if pandemic still/will/... allows it).

benfrain:
Yes. No worries 👍

PR #15, opened 2021-12-08
fix: hide "../" when in root dir

skebanga:
Currently file_browser will always insert an "up dir" entry (../),
regardless of where it is in the filesystem

This change suppresses this entry if we are in the root directory (/)

Note I do not have access to Windows. I don't think \ is ever a
valid path in Windows, so don't believe this will have any adverse
affects on that OS, but I haven't been able to check specifically.

### DIFF
--- a/lua/telescope/_extensions/file_browser/finders.lua
+++ b/lua/telescope/_extensions/file_browser/finders.lua
@@ -32,7 +32,9 @@ fb_finders.browse_files = function(opts)
       table.insert(data, typ == "directory" and (entry .. os_sep) or entry)
     end,
   })
-  table.insert(data, 1, ".." .. os_sep)
+  if opts.path ~= os_sep then
+    table.insert(data, 1, ".." .. os_sep)
+  end
   -- returns copy with properly set cwd for entry maker
   return finders.new_table { results = data, entry_maker = opts.entry_maker { cwd = opts.path } }
 end


fdschmidt93 (commented):
Generally makes sense, however I think there is a slight mix up.

This should be changed in `fb_finders.browse_files` rather than `fb_finders.browse_folders`, so in line 35 we'd need this diff.

```diff
-  table.insert(data, 1, ".." .. os_sep)
+  if opts.cwd ~= os_sep then
+    table.insert(data, 1, ".." .. os_sep)
+  end
```
with the original change of the PR reverted, of course. Or am I confused? :sweat_smile: Could you please confirm & test? :)

> Note I do not have access to Windows. I don't think \ is ever a valid path in Windows, so don't believe this will have any adverse affects on that OS, but I haven't been able to check specifically.

Me neither. Frankly, I'll probably will have the policy that OS-specific issues will have to be fixed by people that use the extension on anything other than Linux. Thanks for pointing this out, reminds me this should be added to the README prior to the file-browser being deprecated upstream :)

However, this maybe also just as well work for Windows, see eg this [SO](https://stackoverflow.com/questions/151860/root-folder-equivalent-in-windows) post.

skebanga:
I'll be honest, I tested it on the master branch of telescope, using the built-in file_browser (see [this now closed PR](https://github.com/nvim-telescope/telescope.nvim/pull/1568)) - I assumed it was a simple copy-paste, but perhaps I was wrong.

Is there a particular branch in [`nvim-telescope/telescope.nvim`](https://github.com/nvim-telescope/telescope.nvim) which disables the built-in file_browser, and replaces it with this repo's file browser?

Alternatively, are there instructions anywhere, or can you give me pointers, on how to actually enable this one and disable the one in the master branch of telescope?

skebanga:
but with that said, yes of course, you are correct that this one is incorrect!

fdschmidt93:
> Alternatively, are there instructions anywhere, or can you give me pointers, on how to actually enable this one and disable the one in the master branch of telescope?

Sorry, but I'm confused: all the instructions are on the README? It's just another extension.

> Is there a particular branch in nvim-telescope/telescope.nvim which disables the built-in file_browser, and replaces it with this repo's file browser?

Not yet, a deprecation notice will be added towards end of this week and the file browser will be removed end of next week.

https://github.com/nvim-telescope/telescope.nvim/issues/1470

lua/telescope/_extensions/file_browser/finders.lua
```diff
@@ -32,7 +32,9 @@ fb_finders.browse_files = function(opts)
       table.insert(data, typ == "directory" and (entry .. os_sep) or entry)
     end,
   })
-  table.insert(data, 1, ".." .. os_sep)
+  if opts.cwd ~= os_sep then
```
fdschmidt93, line 35: Sorry, my bad in the diff. This should be `opts.path`.

skebanga: lol! Sorry - we'll get there! :)

skebanga:
> Sorry, but I'm confused: all the instructions are on the README? It's just another extension.

d'oh - sorry - I was calling the built-in with my keybindings

PR #16, opened 2021-12-08
fix: clarify depth docs and make `false` viable

fdschmidt93:


fdschmidt93:
LGTM thanks!! /s

Just making a such that docs are updated :)

PR #19, opened 2021-12-08
fix: results title on relative path

### LINKED ISSUE #18: file_browser_relative incorrect title on panel
### Description

When use 
```lua
require("telescope").extensions.file_browser.file_browser({
    path = vim.fn.expand('%:h')
  })
```
to invoke the file browser relative to the current path it shows `./` on the title bar
Screen Shot 2021-12-08 at 20 03 11

but if move to another directory and come back show the directory name ok.
Screen Shot 2021-12-08 at 20 04 00

I tried this other option for the path parameter
```lua
  vim.fn.fnamemodify(vim.api.nvim_buf_get_name(0), ':h')
```
but always get the same result

### Neovim version

```markdown
NVIM v0.6.0
Build type: Release
LuaJIT 2.1.0-beta3
```

### Operating system and version

maxOS 11.6.1

### Steps to reproduce

1. `nvim -nu minimal.lua`
2. execute on the command line `lua require("telescope").extensions.file_browser.file_browser({ path = vim.fn.expand('%:h') })`
3. check the panel title

### Expected behavior

show the correct directory path

### Actual behavior

always shows `./`

### Minimal config

```Lua
vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvim/site]]
local package_root = '/tmp/nvim/site/pack'
local install_path = package_root .. '/packer/start/packer.nvim'
local function load_plugins()
  require('packer').startup {
    {
      'wbthomason/packer.nvim',
      {
        'nvim-telescope/telescope.nvim',
        requires = {
          'nvim-lua/plenary.nvim',
          'nvim-telescope/telescope-file-browser.nvim',
        },
      },
      -- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. '/plugin/packer_compiled.lua',
      display = { non_interactive = true },
    },
  }
end
_G.load_config = function()
  require('telescope').setup()
  require('telescope').load_extension('file_browser')
  -- ADD INIT.LUA SETTINGS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
end
if vim.fn.isdirectory(install_path) == 0 then
  print("Installing Telescope and dependencies.")
  vim.fn.system { 'git', 'clone', '--depth=1', 'https://github.com/wbthomason/packer.nvim', install_path }
end
load_plugins()
require('packer').sync()
vim.cmd [[autocmd User PackerComplete ++once echo "Ready!" | lua load_config()]]
```

fdschmidt93:
Ideally fixes #18 

@kascote could you please confirm whether this fixes your issue? The repro steps are incomplete since, if I understand correctly, you need to launch the file browser from a buffer within a sub-directory of `cwd` for the issue to occur as described.

### DIFF
--- a/lua/telescope/_extensions/file_browser/actions.lua
+++ b/lua/telescope/_extensions/file_browser/actions.lua
@@ -384,6 +384,10 @@ fb_actions.goto_cwd = function(prompt_bufnr)
   local finder = current_picker.finder
   finder.path = vim.loop.cwd() .. os_sep
   finder.files = true
+  if current_picker.results_border then
+    local new_title = Path:new(finder.path):make_relative(finder.path) .. os_sep
+    current_picker.results_border:change_title(new_title)
+  end
   current_picker:refresh(false, { reset_prompt = true })
 end
 
--- a/lua/telescope/_extensions/file_browser/picker.lua
+++ b/lua/telescope/_extensions/file_browser/picker.lua
@@ -53,10 +53,11 @@ fb_picker.file_browser = function(opts)
   local cwd = vim.loop.cwd()
   opts.depth = vim.F.if_nil(opts.depth, 1)
   opts.cwd = opts.cwd and vim.fn.expand(opts.cwd) or cwd
+  opts.path = vim.F.if_nil(opts.path, cwd)
   opts.files = vim.F.if_nil(opts.files, true)
   pickers.new(opts, {
     prompt_title = opts.files and "File Browser" or "Folder Browser",
-    results_title = opts.files and Path:new(opts.cwd):make_relative(cwd) .. os_sep or "Results",
+    results_title = opts.files and Path:new(opts.path):make_relative(cwd) .. os_sep or "Results",
     finder = fb_finder.finder(opts),
     previewer = conf.file_previewer(opts),
     sorter = conf.file_sorter(opts),


kascote:
yesssss, thanks you !

fdschmidt93:
Thanks for the quick testing :)

ikerurda:
With the theme ivy this is not fixed. In fact, it's only shown when in "./", otherwise is hidden

fdschmidt93:
I don't follow. ivy theme doesn't show a results title for me? How could this be an issue then. I cannot reproduce. Please full repro steps.

ikerurda:
Screenshot 2021-12-09 at 1 03 04 AM

Just with this minimal setup, **when in cwd** "./" is shown down there. Then doesn't matter which folder I enter, it goes away
```lua
require("telescope").setup {
	extensions = {
		file_browser = {
			theme = "ivy",
			layout_config = {height = .5}
		}
	}
}
```
This doen't happen with other themes... no idea

fdschmidt93:
>  when in cwd "./" is shown down there

This is expected behavior for the results title as such.

As to why it shows up for you and not for me, I guess the height parameter makes the difference. If you enter another folder it should show eg "after" when entering after. "./" just indicates you are in cwd.

PR #20, opened 2021-12-09
feat: change directory based on spcieal input paths

skebanga:
This change adds the following 2 special actions based on the input
path:

- If the user enters `~/` in the TelescopePrompt buffer, the
  file_browser is automatically changed to the user's `${HOME}`
directory

- If the user enters `/` in the TelescopePrompt buffer, the file_browser
  is automatically changed to the root directory (`/`)

The latter action (switching to root) clashes with the existing behaviour
where entering `/` into the TelescopePrompt buffer switches to the directory
browser.

Since this behaviour is also available under `<C-f>` (in insert mode) and
`f` (in normal mode), losing this special action was deemed acceptable.

### DIFF
--- a/lua/telescope/_extensions/file_browser.lua
+++ b/lua/telescope/_extensions/file_browser.lua
@@ -85,15 +85,12 @@ local pconf = {
     },
   },
   on_input_filter_cb = function(prompt)
-    if prompt:sub(-1, -1) == os_sep then
+    if prompt == '~/' then
       local prompt_bufnr = vim.api.nvim_get_current_buf()
-      if vim.bo[prompt_bufnr].filetype == "TelescopePrompt" then
-        local current_picker = action_state.get_current_picker(prompt_bufnr)
-        if current_picker.finder.files then
-          fb_actions.toggle_browser(prompt_bufnr, { reset_prompt = true })
-          current_picker:set_prompt(prompt:sub(1, -2))
-        end
-      end
+      fb_actions.goto_dir(prompt_bufnr, vim.env.HOME .. os_sep)
+    elseif prompt == '/' then
+      local prompt_bufnr = vim.api.nvim_get_current_buf()
+      fb_actions.goto_dir(prompt_bufnr, prompt)
     end
   end,
   attach_mappings = function(prompt_bufnr, _)
--- a/lua/telescope/_extensions/file_browser/actions.lua
+++ b/lua/telescope/_extensions/file_browser/actions.lua
@@ -356,6 +356,17 @@ fb_actions.open_file = function(prompt_bufnr)
   actions.close(prompt_bufnr)
 end
 
+--- Goto directory in |builtin.file_browser|.
+---@param prompt_bufnr number: The prompt bufnr
+---@param path string: Directory to go to
+fb_actions.goto_dir = function(prompt_bufnr, path)
+  local current_picker = action_state.get_current_picker(prompt_bufnr)
+  local finder = current_picker.finder
+  finder.path = path
+  finder.files = true
+  current_picker:refresh(false, { reset_prompt = true })
+end
+
 --- Goto parent directory in |builtin.file_browser|.
 ---@param prompt_bufnr number: The prompt bufnr
 ---@param bypass boolean: Allow passing beyond the globally set current working directory
@@ -363,7 +374,6 @@ fb_actions.goto_parent_dir = function(prompt_bufnr, bypass)
   bypass = vim.F.if_nil(bypass, true)
   local current_picker = action_state.get_current_picker(prompt_bufnr)
   local finder = current_picker.finder
-  local parent_dir = Path:new(finder.path):parent()
 
   if not bypass then
     if vim.loop.cwd() == finder.path then
@@ -372,9 +382,8 @@ fb_actions.goto_parent_dir = function(prompt_bufnr, bypass)
     end
   end
 
-  finder.path = parent_dir .. os_sep
-  finder.files = true
-  current_picker:refresh(false, { reset_prompt = true })
+  local path = Path:new(finder.path):parent() .. os_sep
+  fb_actions.goto_dir(prompt_bufnr, path)
 end
 
 --- Goto working directory of nvim in |builtin.file_browser|.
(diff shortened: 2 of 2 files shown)


lua/telescope/_extensions/file_browser.lua
```diff
@@ -85,15 +85,12 @@ local pconf = {
     },
   },
   on_input_filter_cb = function(prompt)
-    if prompt:sub(-1, -1) == os_sep then
+    if prompt == '~/' then
```
fdschmidt93, line 88: I'm generally open to having such input callbacks as users can unset them, but we'd have to make sure, that they are reasonably OS-agnostic, as in entering `~/` or `/` should not trigger errors on Windows (should work as is on macOS, I presume).

lua/telescope/_extensions/file_browser/actions.lua
```diff
@@ -356,14 +356,24 @@ fb_actions.open_file = function(prompt_bufnr)
   actions.close(prompt_bufnr)
 end
 
+--- Goto directory in |builtin.file_browser|.
+---@param prompt_bufnr number: The prompt bufnr
+---@param path string: Directory to go to
+fb_actions.goto_dir = function(prompt_bufnr, path)
```
fdschmidt93, line 362: I guess this abstraction generally makes sense, but lua does not have partial functions by default. I'm not opposed to having this, but the docs for the function should at a code example on how to use this for own mappings.

telescope core repo has a lot of code tags for reference.

fdschmidt93 (commented):
Generally LGTM, I haven't tried it (will do soonish, say before telescope fb deprecation).

I think it's cool, but I kind of would like to have broader user input on such a PR, because it can always be turned opt-in by setting `on_input_cb` in the extension setup I guess as opposed to default behavior. Thus, you could make the case this PR could just as well be a recipe similar to what we have in telescope.

I'll try to think about it more because I guess user input won't come :sweat_smile:

skebanga:
> it can always be turned opt-in by setting `on_input_cb` in the extension setup I guess as opposed to default behavior

I wasn't aware of this. I need to RTFM a bit more! (Actually, just went to try read up on it and I see it's not documented, so maybe I can be forgiven! :grin:)

I think you're right, though, that this probably should just be a recipe...

lua/telescope/_extensions/file_browser.lua
skebanga: True - `/` would likely cause an issue on Windows - more evidence this should probably be a recipe

lua/telescope/_extensions/file_browser/actions.lua
skebanga: Please can you elaborate? I'm a total noob when it comes to lua, so there are definitely tonnes of nuances I'm not currently aware of.

However, with that said, this isn't a partial function is it? It's just a normal function which takes a path, and sets the finder path to that.

So it's used in the original intent for this PR, which is to call `fb_actions.goto_dir(prompt_bufnr, "~/")` when the user types `"~/"`, and similarly it calls `fb_actions.goto_dir(prompt_bufnr, "/")` when the user types `"/"`.

However, since there seemed to be some duplication in the other two functions below (`fb_actions.goto_parent_dir` and fb_actions.goto_cwd`), instead of having that duplication I just call `fb_actions.goto_dir` with parent dir and the cwd respectively.

fdschmidt93: > Please can you elaborate? I'm a total noob when it comes to lua, so there are definitely tonnes of nuances I'm not currently aware of.

> However, with that said, this isn't a partial function is it? It's just a normal function which takes a path, and sets the finder path to that.

The function couldn't be mapped straightforwardly, because you'd have to wrap it in another function. Hence, it's not really an action as such but more of a helper.

skebanga: Ah ok, I see.

skebanga:
I'm struggling to work out how to set my own `on_input_filter_cb`

This is what I tried:

```lua
require('telescope').setup {
  extensions = {
    file_browser = {
      on_input_filter_cb = function(prompt)
        -- check prompt here
        end
    }
  }
}
```

This doesn't seem to work. Is this the right way to do it?

fdschmidt93:
Seems to work for me. The below overrides the default input cb and pretty much always prints "test".

```lua
require "telescope".setup {
  extensions = {
    file_browser = {
      on_input_filter_cb = function()
        print "test"
      end,
    },
  },
}
```

Overrides the existing input callbacks

skebanga:
That's super weird - I tried to copy your code snippet and it doesn't print "test" at all! Not sure where I'm going wrong! :(

skebanga:
Ah, finally got it working! 

In case it's of use to anyone in the future, here is my code...

**`lua/config/telescope.lua`:**
```lua
local M = {}

function M.setup()
  local telescope = require("telescope")
  local action_state = require("telescope.actions.state")

  local file_browser_change_dir = function(path)
    local prompt_bufnr = vim.api.nvim_get_current_buf()
    finder.path = path
    finder.files = true
    current_picker:refresh(false, { reset_prompt = true })
  end

  telescope.setup({
    extensions = {
      file_browser = {
        on_input_filter_cb = function(prompt)
          if prompt == '~/' then
            file_browser_change_dir(vim.env.HOME .. '/')
          elseif prompt == '/' then
            file_browser_change_dir('/')
          end
        end,
      },
    },
  })

  telescope.load_extension "file_browser"
end

return M
```

In my `packer` config I set the `config` to be the above `setup` function

```lua
use({
  "nvim-telescope/telescope.nvim",
  requires = {
    {"nvim-telescope/telescope-file-browser.nvim"},
  },
  config = require("config.telescope").setup,
})
```

fdschmidt93:
You're very happy to add a version of this PR to https://github.com/nvim-telescope/telescope-file-browser.nvim/wiki/Configuration-Recipes :)

skebanga:
Doesn't seem possible to fork a wiki

https://github.community/t/how-to-fork-a-wiki-and-send-a-pr/2006

fdschmidt93:
I think you can just edit this wiki here? Maybe I have to change some settings

PR #22, opened 2021-12-11
fix: batch rename borderchars and calceling rename

ikerurda:
`require("telescope.config").values.borderchars` returns a table with borderchars for preview, prompt and results, not some borderchars.

canceling `actions.rename_file()` was missing `return`

### DIFF
--- a/lua/telescope/_extensions/file_browser/actions.lua
+++ b/lua/telescope/_extensions/file_browser/actions.lua
@@ -104,7 +104,7 @@ local batch_rename = function(prompt_bufnr, selections)
     enter = true,
     noautocmd = true,
     border = { 1, 1, 1, 1 },
-    borderchars = config.values.borderchars.preview,
+    borderchars = config.values.borderchars,
   }
   local win, win_opts = popup.create(buf, popup_opts)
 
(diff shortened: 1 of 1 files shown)


fdschmidt93:
> canceling actions.rename_file() was missing return

Thanks for catching that :)

> require("telescope.config").values.borderchars returns a table with borderchars for preview, prompt and results, not some borderchars.

Can you please elaborate when exactly this is an issue? https://github.com/nvim-telescope/telescope.nvim/blob/5f37fbfa837dfee7ecd30f388b271f4a71c0a9e0/lua/telescope/config.lua#L323

`:lua print(vim.inspect(require("telescope.config").values.borderchars))` just returns a table with borderchars for me on telescope master and not anything with preview, prompt, and results

ikerurda:
True, sorry, my bad. I was setting my default theme to dropdown:
```lua
require("telescope").setup {
    defaults = require("telescope.themes").get_dropdown {...},
    ...
}
```
Should we consider this a bug or are we supposed not to change the default theme that way?

fdschmidt93:
> Should we consider this a bug or are we supposed not to change the default theme that way?

I'd prefer to leave it as is as the theme should not be set that way, since it's unexpectedly changing the telescope configuration. In other words, the cleanest fix should probably stem from an upstream telescope change.

The issue here also is that there may or may not be a preview. Hence, your proposed solution is not universal enough to accept it as a current workaround.

ikerurda:
Understood. Thanks! :)

fdschmidt93:
Thanks :)

Issue #31, opened 2021-12-20
Error when border = false


PaterJason:
### Description

The picker crashes when border = false

I have a potential fix on my branch
https://github.com/PaterJason/telescope-file-browser.nvim/commit/1ab67bc7d3a97f008992ff4f335313fb02345648

### Neovim version

```markdown
NVIM v0.6.0
Build type: Release
LuaJIT 2.0.5
```

### Operating system and version

Arch Linux

### Steps to reproduce

1. `nvim -nu minimal.lua`
2. Command `lua require 'telescope'.extensions.file_browser.file_browser()<CR>`
3. Enter any directory

### Actual behavior

```
E5108: Error executing lua ...-browser.nvim/lua/telescope/_extensions/file_browser.lua:96: attempt to index field 'prompt_border' (a nil value)
stack traceback:
        ...-browser.nvim/lua/telescope/_extensions/file_browser.lua:96: in function 'run_replace_or_original'
        ...packer/start/telescope.nvim/lua/telescope/actions/mt.lua:30: in function 'run_replace_or_original'
        ...packer/start/telescope.nvim/lua/telescope/actions/mt.lua:30: in function 'key_func'
        ...k/packer/start/telescope.nvim/lua/telescope/mappings.lua:235: in function 'execute_keymap'
        [string ":lua"]:1: in main chunk
```

### Minimal config

```Lua
vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvim/site]]
local package_root = '/tmp/nvim/site/pack'
local install_path = package_root .. '/packer/start/packer.nvim'
local function load_plugins()
  require('packer').startup {
    {
      'wbthomason/packer.nvim',
      {
        'nvim-telescope/telescope.nvim',
        requires = {
          'nvim-lua/plenary.nvim',
          'nvim-telescope/telescope-file-browser.nvim',
        },
      },
      -- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. '/plugin/packer_compiled.lua',
      display = { non_interactive = true },
    },
  }
end
_G.load_config = function()
  require('telescope').setup{ defaults = { border = false } }
  require('telescope').load_extension('file_browser')
  -- ADD INIT.LUA SETTINGS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
end
[...]

PR #32, opened 2021-12-20
fix: border = false error

PaterJason:
#31

When `defaults = { border = false }` selecting any directory results in an error


fdschmidt93 (commented):
LGTM thanks :) 

Just as a side note, the border titles (only shown with a border) comprises valuable information wrt to the file browser, but you most likely are aware and don't mind :sweat_smile:

PR #35, opened 2021-12-28
feat: fd file and folder finders 

fdschmidt93:
Round 2 of #26

(See rationale for `fd` on original PR)


fdschmidt93:
I think this is fine this time around :) fingers crossed :sweat_smile:

Issue #37, opened 2021-12-29
Incorrect file copied message when a file was actually not copy


miguelmoraperea:
### Description

Currently telescope-file-browser.nvim does not support copying a file to another directory other than cwd.

If a user tries to copy a file to another directory, the file is not copied, but the user gets the following message: 

`/example/path/to/file has been copied!`

Which is an incorrect and misleading message, the file in fact was never copied.

### Neovim version

```markdown
NVIM v0.7.0-dev+792-gc46f7caad
Build type: RelWithDebInfo
```

### Operating system and version

Ubuntu 20.04

### Steps to reproduce

1. Start to copy a file
2. When prompt: "Copying existing file or folder within original directory, please provide a new file or folder name:" edit the file path to another directory, other than cwd.
3. The file will not be copied (as expected) but the user will get the following message "[file] has been copied!".

### Expected behavior

When copying a file outside the original directory, the message to the user should be:  "[file] has not been copied".

### Actual behavior

When copying a file outside the original directory, the message currently is:  "[file] has been copied!", which is incorrect.

### Minimal config

```Lua
vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvim/site]]
local package_root = '/tmp/nvim/site/pack'
local install_path = package_root .. '/packer/start/packer.nvim'
local function load_plugins()
  require('packer').startup {
    {
      'wbthomason/packer.nvim',
      {
        'nvim-telescope/telescope.nvim',
        requires = {
          'nvim-lua/plenary.nvim',
          'nvim-telescope/telescope-file-browser.nvim',
        },
      },
      -- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. '/plugin/packer_compiled.lua',
      display = { non_interactive = true },
    },
  }
end
_G.load_config = function()
  require('telescope').setup()
  require('telescope').load_extension('file_browser')
  -- ADD INIT.LUA SETTINGS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
end
if vim.fn.isdirectory(install_path) == 0 then
  print("Installing Telescope and dependencies.")
  vim.fn.system { 'git', 'clone', '--depth=1', 'https://github.com/wbthomason/packer.nvim', install_path }
end
load_plugins()
require('packer').sync()
[...]

PR #39, opened 2021-12-29
fix: Incorrect file copied message (#37)

miguelmoraperea:
When copying a file outside the original directory (cwd) the file is not
copied, this is expected, but the user still gets the following message:

  [file] has been copied!

This is incorrect, this commit fixes the issue by verifying the result
status from the copy operation and prints the appropriate message
'[file] has been copied!' or '[file] has not been copied' depending on
whether the operation succeeded or not.


fdschmidt93:
Thanks for your PR :) Irrespective of the discussion on the original issue, this seems like a good change.

I'll try to test drive it and look into it soon.

miguelmoraperea:
I might've moved some branches in my forked repo, and it seems this PR got closed automatically, anyways I opened a new one: 
https://github.com/nvim-telescope/telescope-file-browser.nvim/pull/45

:)

Issue #3, opened 2021-11-14
Roadmap


fdschmidt93:
This issue will keep users informed of any roadmap, gotchas, breaking changes, and the like.

# Roadmap

The next focus will be on refactoring 

* the entry maker and associated path display
* previewer (maybe with upstream PRs) to facilitate customization

## TODOs (PRs welcome :)

TODOs are not really sorted by priority.

- [ ] A view of multi-selections (cf this [comment](https://github.com/nvim-telescope/telescope-file-browser.nvim/pull/41#issuecomment-1004624979))
- [ ] **Warning**: `Batch Rename` (renaming multi-selections) does not account for links between renamed objects yet (difficult fix). In any case, it should be safe to rename disjoint paths.
- [ ] Better handle moving files/folders to targets in which files/folders already exist, see "skipping" in [fb_actions](https://github.com/nvim-telescope/telescope-file-browser.nvim/blob/master/lua/telescope/_extensions/file_browser/actions.lua)
- [ ] Being able to toggle `depth` in the file browser as `false` is unlimited depth, could be interesting
- [ ] Maybe allow to set application by extension with which a file is opened
- [ ] Own previewer for making previewer truly customize-able (will happen upstream)

The aforementioned list will be continuously updated.

# Contributions

Any help on the above mentioned issues is greatly appreciated! To get started, familiarize with the code base and [plenary's pathlib](https://github.com/nvim-lua/plenary.nvim/blob/master/lua/plenary/path.lua). In fact, I'm very much welcome a main maintainer of the project as my work on telescope core already takes a lot of my time allocated to open source.

fdschmidt93:
Important notice: the `Batch Rename` should be used with paths that have interdependence.

For instance, you select a folder and then a sub-file or sub-folder to rename. Since there is no path resolution right now, this fails for the latter paths as they have become invalid. This is non-trivial to resolve.

Nevertheless, atomically renaming multiple files and folders without interdependence should work fine.

fdschmidt93:
As an update: apart from deprecating the file-browser upstream and maybe cleaning up remappings (in co-op with upstream PR), I'll unfortunately won't have too much time up until end of February as work keeps me preoccupied until end of this year and I'll (hopefully) will be travelling throughout January.

With 0.6 now the basic functionality should be working reasonably well and I'll do my best to attend to the small issues in the meantime, too.

Once more, I'd be very happy to get other people on board to help out with developing the extension :)

[...]

PR #41, opened 2021-12-29
fix: ignore '..' and '.' during select/toggle all

jamestrew:
This should check off the third item in the med priority list in #3.

I didn't see any instances of `./` showing up anywhere so I didn't really consider them. Really only looked at preventing `../` from being selected.

I also left references to `../` and `./` in the `ignore_values` but I can remove them if needed.

### DIFF
--- a/lua/telescope/_extensions/file_browser/actions.lua
+++ b/lua/telescope/_extensions/file_browser/actions.lua
@@ -437,19 +437,14 @@ fb_actions.toggle_browser = function(prompt_bufnr, opts)
   current_picker:refresh(finder, { reset_prompt = opts.reset_prompt, multi = current_picker._multi })
 end
 
-local function _get_ignore_values(current_picker)
-  local finder = current_picker.finder
-  local path_parent = tostring(Path:new(finder.path):parent())
-  return { "../", "./", path_parent, finder.path }
-end
-
 --- Toggles all selections akin to |actions.toggle_all| but ignores parent & current directory
 ---@param prompt_bufnr number: The prompt bufnr
 fb_actions.toggle_all = function(prompt_bufnr)
   local current_picker = action_state.get_current_picker(prompt_bufnr)
-  local ignore_values = _get_ignore_values(current_picker)
+  local finder = current_picker.finder
+  local parent_dir = tostring(Path:new(finder.path):parent())
   action_utils.map_entries(prompt_bufnr, function(entry, _, row)
-    if not vim.tbl_contains(ignore_values, entry.value) then
+    if not vim.tbl_contains({ finder.path, parent_dir }, entry.value) then
       current_picker._multi:toggle(entry)
       if current_picker:can_select_row(row) then
         current_picker.highlighter:hi_multiselect(row, current_picker._multi:is_selected(entry))
@@ -463,10 +458,11 @@ end
 ---@param prompt_bufnr number: The prompt bufnr
 fb_actions.select_all = function(prompt_bufnr)
   local current_picker = action_state.get_current_picker(prompt_bufnr)
-  local ignore_values = _get_ignore_values(current_picker)
+  local finder = current_picker.finder
+  local parent_dir = tostring(Path:new(finder.path):parent())
   action_utils.map_entries(prompt_bufnr, function(entry, _, row)
     if not current_picker._multi:is_selected(entry) then
-      if not vim.tbl_contains(ignore_values, entry.value) then
+      if not vim.tbl_contains({ finder.path, parent_dir }, entry.value) then
         current_picker._multi:add(entry)
         if current_picker:can_select_row(row) then
           current_picker.highlighter:hi_multiselect(row, current_picker._multi:is_selected(entry))
--- a/lua/telescope/_extensions/file_browser/actions.lua
+++ b/lua/telescope/_extensions/file_browser/actions.lua
@@ -165,6 +165,7 @@ end
 fb_actions.rename_file = function(prompt_bufnr)
   local current_picker = action_state.get_current_picker(prompt_bufnr)
   local selections = fb_utils.get_selected_files(prompt_bufnr, false)
+  local parent_dir = Path:new(current_picker.finder.path):parent()
 
   if not vim.tbl_isempty(selections) then
     batch_rename(prompt_bufnr, selections)
@@ -175,8 +176,8 @@ fb_actions.rename_file = function(prompt_bufnr)
       return
     end
     local old_path = Path:new(entry[1])
-    -- "../" more common so test first
-    if old_path.filename == "../" or old_path.filename == "./" then
+    -- "../" aka parent_dir more common so test first
+    if old_path.filename == parent_dir.filename then
       print "Please select a file!"
       return
     end
@@ -185,26 +186,26 @@ fb_actions.rename_file = function(prompt_bufnr)
       print "Renaming file aborted."
       return
     end
-    new_name = Path:new(new_name)
+    local new_path = Path:new(new_name)
 
-    if old_path.filename == new_name.filename then
+    if old_path.filename == new_path.filename then
       print "Original and new filename are the same! Skipping."
       return
     end
 
-    if new_name:exists() then
-      print(string.format("%s already exists! Skipping.", new_name.filename))
+    if new_path:exists() then
+      print(string.format("%s already exists! Skipping.", new_path.filename))
       return
     end
 
     -- rename changes old_name in place
     local old_name = old_path:absolute()
 
-    old_path:rename { new_name = new_name.filename }
-    if not new_name:is_dir() then
-      fb_utils.rename_buf(old_name, new_name:absolute())
+    old_path:rename { new_name = new_path.filename }
+    if not new_path:is_dir() then
+      fb_utils.rename_buf(old_name, new_path:absolute())
     else
-      fb_utils.rename_dir_buf(old_name, new_name:absolute())
+      fb_utils.rename_dir_buf(old_name, new_path:absolute())
     end
 
     -- persist multi selections unambiguously by only removing renamed entry
@@ -443,7 +444,7 @@ end
 fb_actions.toggle_all = function(prompt_bufnr)
   local current_picker = action_state.get_current_picker(prompt_bufnr)
   local finder = current_picker.finder
-  local parent_dir = tostring(Path:new(finder.path):parent())
+  local parent_dir = Path:new(finder.path):parent().filename
   action_utils.map_entries(prompt_bufnr, function(entry, _, row)
     if not vim.tbl_contains({ finder.path, parent_dir }, entry.value) then
       current_picker._multi:toggle(entry)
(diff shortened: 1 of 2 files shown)


jamestrew:
After some more testing, I'm not satisfied with the behavior of my initial implementation, particularly when using the `folder_browser`. I also realized the `./` becomes relevant when it comes to the `folder_browser`. Going to do some tweaking.

jamestrew:
Without further changing the behavior of `toggle_all` and `select_all` to clear already selected files, this sort of thing happens when you `toggle_all` then navigate lower into the directories where `../` is ignored but ignored while selected.
image

Personally, I find it a little unintuitive for selections to persist through multiple directories when you can only really view one directory at any given time. Maybe it's worth considering changing selections to be reset when moving to a different directory however this would impact the `move_file` and `copy_file` functionality. Thoughts?

lua/telescope/_extensions/file_browser/actions.lua
```diff
@@ -437,11 +437,17 @@ fb_actions.toggle_browser = function(prompt_bufnr, opts)
...
+end
+
+--- Toggles all selections akin to |actions.toggle_all| but ignores parent & current directory
 ---@param prompt_bufnr number: The prompt bufnr
 fb_actions.toggle_all = function(prompt_bufnr)
   local current_picker = action_state.get_current_picker(prompt_bufnr)
-  local ignore_values = { "../", "./" }
+  local ignore_values = _get_ignore_values(current_picker)
   action_utils.map_entries(prompt_bufnr, function(entry, _, row)
     if not vim.tbl_contains(ignore_values, entry.value) then
```
fdschmidt93, line 452: The changes can probably be reduced to

```lua
    local parent_dir = Path:new(finder.path):parent():absolute()
    if not vim.tbl_contains({finder.path, parent_dir}, entry.value) then
```

Of course, analogously down below.

fdschmidt93 (commented):
> this sort of thing happens when you toggle_all then navigate lower into the directories where ../ is ignored but ignored while selected.

Your gif is a bit weird to follow because without border titles it is unclear in what directory you are triggering selection. In any case, `..` and `.` (and their selections) refer to absolute paths which of course change whenever you change the directory. Isn't that what you are observing? Apologies, can you please reiterate what's confusing?

> Personally, I find it a little unintuitive for selections to persist through multiple directories when you can only really view one directory at any given time.

That will definitely not change since it is one of the primary features and selling points of `telescope-file-browser`. Entries are cached by their absolute paths to resolve multi selections across file and folder finder and varying directories, enabling swift copying, removing, and renaming files across folders. What may be a worthwhile feature to add (also for telescope) is having a some sort view of currently selected entries.

jamestrew:
Here's a new gif with the borders. There seems to be a separate issue with the `ivy` theme causing the `results_title` to bug out. My colorscheme makes the `results_title` hard to see but initially I see `./` as I should, but when I start navigating into folders, `results_title` seems to just blank-out. Switching to `browse_folder` still shows nothing. I will try to look into this separately.
image

Gif again with the default theme with screenkey so you can better tell what I'm doing:
image
Here you can (barely) see the `results_title` working correctly.

So I have a couple of observation/questions:
Firstly, the topic of this PR, my current implementation ignores the `finder.path` and the parent of `finder.path`. My concern is when a user selects all files then proceeds to navigates around, `finder.path` changes as it should, and the corresponding `ignore_values` changes, causing `../` (the parent path) to stay selected in instances where it was previously selected. This is demonstrated in the gif when I select all, then navigate to `./telescop-file-browser/lua/telescope` and toggle all. Is this your desired behavior? Or would you like to always have `../` & `./` unselected when selecting/toggling all?

Second, (and this is probably another separate issue but I wanted to mention it since it occurs in the second gif) when I toggle all on/off, some selections don't get unselected. In the second gif, I toggle selections on and eventually return back to the project root directory, then toggle selections off. But `/scripts/` stays selected (at least visually) until I start moving my "selector". I can only reproduce it consistently in this `telescope-file-browser` directory and it's always the `/scripts/` folder... I will try to look into that as well.

fdschmidt93:
> Here's a new gif with the borders. There seems to be a separate issue with the ivy theme causing the results_title to bug out. My colorscheme makes the results_title hard to see but initially I see ./ as I should, but when I start navigating into folders, results_title seems to just blank-out. Switching to browse_folder still shows nothing. I will try to look into this separately.

This pretty surely is an upstream `telescope.nvim` issue and cannot be solved from within `telescope-file-browser.nvim`. Results titles was only added long after `ivy` theme. I guess something doesn't play well. Very welcome for you to look into this :)

fdschmidt93:
> Second, (and this is probably another separate issue but I wanted to mention it since it occurs in the second gif) when I toggle all on/off, some selections don't get unselected. In the second gif, I toggle selections on and eventually return back to the project root directory, then toggle selections off. But /scripts/ stays selected (at least visually) until I start moving my "selector". I can only reproduce it consistently in this telescope-file-browser directory and it's always the /scripts/ folder... I will try to look into that as well.

I'm quite certain that's only a visual issue, I can eg also reproduce this with `builtin.find_files`. You can confirm this with (it's a bit complicated, I have short hands for these things :sweat_smile:):

```lua
lua TelescopeGlobalState[vim.api.nvim_get_current_buffer()].picker._multi._entries
```

should be empty.

fdschmidt93:
> Firstly, the topic of this PR, my current implementation ignores the finder.path and the parent of finder.path. My concern is when a user selects all files then proceeds to navigates around, finder.path changes as it should, and the corresponding ignore_values changes, causing ../ (the parent path) to stay selected in instances where it was previously selected. This is demonstrated in the gif when I select all, then navigate to ./telescope-file-browser/lua/telescope and toggle all. Is this your desired behavior? Or would you like to always have ../ & ./ unselected when selecting/toggling all?

I'd say that's fine, since 
* it's consistent behavior
* Single items can easily be untoggled with `TAB`
* There's always corresponding [telescope-actions](https://github.com/nvim-telescope/telescope.nvim/blob/a01ebd2793999c11d727fd15b1e5979ba20c7503/lua/telescope/actions/init.lua#L107-L145) users can rely upon if they want to select/drop/toggle all. (**)

We can maybe add a cautionary note to the docs though?

The primary purpose of these `fb_actions` was to introduce ways to more easily select almost all files in a folder for moving or copying files.

(**) reminds me that dropping all entries here acts a bit differently since we have entries across multiple finders. We'd need an action for that to I suppose or change the upstream implementation.. after I implemented it, I guess it makes more sense to change the upstream implementation. Leaving this as reference for people. I 

```lua
--- Drop all multi-selected entries across all folders.
--- - Note: dropped multi-selected entries may include results not visible in the results popup.
---@param prompt_bufnr number: The prompt bufnr
fb_actions.drop_all = function(prompt_bufnr)
  current_picker._multi._entries = {}
  local num_results = math.min(current_picker.max_results, current_picker.manager:num_results())
  for index = 1, num_results do
    local row = current_picker:get_row(index)
    if current_picker:can_select_row(row) then
      current_picker.highlighter:hi_multiselect(row, false)
    end
  end
end
```

jamestrew:
> > Here's a new gif with the borders. There seems to be a separate issue with the ivy theme causing the results_title to bug out. My colorscheme makes the results_title hard to see but initially I see ./ as I should, but when I start navigating into folders, results_title seems to just blank-out. Switching to browse_folder still shows nothing. I will try to look into this separately.
> 
> This pretty surely is an upstream `telescope.nvim` issue and cannot be solved from within `telescope-file-browser.nvim`. Results titles was only added long after `ivy` theme. I guess something doesn't play well. Very welcome for you to look into this :)

I traced the issue to be in `plenary.nvim` so I put in a [PR](https://github.com/nvim-lua/plenary.nvim/pull/300) there. Won't be needing any changes to either `telescope.nvim` or `telescope-file-browser`.

jamestrew:
> > Second, (and this is probably another separate issue but I wanted to mention it since it occurs in the second gif) when I toggle all on/off, some selections don't get unselected. In the second gif, I toggle selections on and eventually return back to the project root directory, then toggle selections off. But /scripts/ stays selected (at least visually) until I start moving my "selector". I can only reproduce it consistently in this telescope-file-browser directory and it's always the /scripts/ folder... I will try to look into that as well.
> 
> I'm quite certain that's only a visual issue, I can eg also reproduce this with `builtin.find_files`. You can confirm this with (it's a bit complicated, I have short hands for these things sweat_smile):
> 
> ```lua
> lua TelescopeGlobalState[vim.api.nvim_get_current_buffer()].picker._multi._entries
> ```
> 
> should be empty in the `script` highlighted-only case.

I can't reproduce this at all anymore so unless more people start raising issues related to this, I'm going to pretend I didn't see it :hand_over_mouth: 

I'll be keeping on eye out on the issues for this repo and helping where I can. I'm still a fairly new software engineer with no formal education in the field so I'll apologize & thank you in advance for any shenanigans on my part and patience on your part.

jamestrew:
I added a little bit more to the docstrings of the two actions. The docs will be auto generated during the merge correct?

fdschmidt93:
> I'll be keeping on eye out on the issues for this repo and helping where I can. I'm still a fairly new software engineer with no formal education in the field so I'll apologize & thank you in advance for any shenanigans on my part and patience on your part.

No worries :) same here. 

> I added a little bit more to the docstrings of the two actions. The docs will be auto generated during the merge correct?

They should already have been added to the PR as a separate commit. I think that happens with first-time contributors because the workflow is not fully activated from the get go. I'll look into it.

Sleeping over this once more after our continued discussion I better grokked

> Personally, I find it a little unintuitive for selections to persist through multiple directories when you can only really view one directory at any given time.

that multi-selections behave quite differently from vanilla `telescope.nvim` pickers (cf. also toggling telescope vs `fb_actions`). I'll add this in vimdocs [here](https://github.com/nvim-telescope/telescope-file-browser.nvim/blob/master/lua/telescope/_extensions/file_browser.lua#L1-L45) and to the README once I'll revise the docs.

Further, I envision that we have some means of inspecting multi-selections something like the following:

- The preview buffer now lists the multi-selections
- The results buffer has a disclaimer float over it, stating that multi-selections are currently shown
- Scrolling is still working
- Some keybinds for toggling multi-selections

This should be rather straightforward with lua function keymaps merged in `nvim` master. Probably will have to be master only up until 0.7.

What do you think?

PR LGTM thanks :) I'll make one more sense check pass and check how to trigger docs asap (probably tomorrow evening, this might turn into a bit of a goose chase).

jamestrew:
> Further, I envision that we have some means of inspecting multi-selections something like the following:
> 
> * The preview buffer now lists the multi-selections
> * The results buffer has a disclaimer float over it, stating that multi-selections are currently shown
> * Scrolling is still working
> * Some keybinds for toggling multi-selections

I like this idea and I think it would reduce the confusion around what's been already selected

jamestrew:
Last thing, I noticed `rename_file` was still using relative paths so I tweak that as well. 
Also very nitpick-y but sumneko lua lsp was struggling with the handling of `new_name` as a string then switching to a `Path` object/table so it was throwing diagnostic errors everywhere.

fdschmidt93:
I think you can now trigger the doc gen under Actions -> Generate Docs -> Run workflow? It doesn't seem to be available for me but it should be there. If not, please try rebasing.

Once docs are generated, I'll merge :sweat_smile:

Issue #13, opened 2021-12-07
How to remove files without prompt?


rzksobhy78:
how can I remove files without this prompt

or at least make it easier by just pressing y for yes or n no
and no additional need to press one more enter

![Tue Dec  7 22:05:37 EET 2021](https://user-images.githubusercontent.com/94202082/145098500-58a436dc-b1e7-422f-8007-d91ee38db2f3.png)

fdschmidt93:
You cannot change that with the default implementation, as that one relies on `vim.fn.inputlist` which produces the kind of selection via numbers as you see in your prompt.

Consequently, the most straightforward answer would be to use a different custom handler for `vim.ui.select` eg https://github.com/nvim-telescope/telescope-ui-select.nvim

However, `telescope-ui-select.nvim` does not block, which might cause issues in case you customize LSP handlers that require blocking up until you confirm the selection.

I'll leave the issue open as I'll reconsider this prior to file browser deprecation in telescope-core, though I'm a bit reluctant to switch the code back to `vim.fn.confirm`

miguelmoraperea:
@rzksobhy78 @fdschmidt93 I agree that using numbered options is not ideal, considering there are only 2 options (yes/no) there will never be more, which is why I also prefer the more user-friendly (Y/N) format.

Do you mind testing my latest pull request, that addresses this request, it works like this:

![refactor-delete-file-options](https://user-images.githubusercontent.com/42681212/147721675-fd63f218-6071-4965-a477-8d0b49e1ac4a.gif)

fdschmidt93:
I think this is now sufficiently addressed with #42 

Please let me know in case there's more to do here.

PR #42, opened 2021-12-30
rfc: Remove selected files user options (#13)

miguelmoraperea:
This commit refactors the user options from the previous style:

  Remove Selected Files:
  1: Yes
  2: No
  Type number and <Enter> or click with the mouse (q or empty cancels):

To a more user-friendly and intuitive format:

  Remove selected files (Y/N):

Where the user can enter 'Y' or 'y' for 'Yes' and any other character
for 'No'.

### DIFF
--- a/lua/telescope/_extensions/file_browser/actions.lua
+++ b/lua/telescope/_extensions/file_browser/actions.lua
@@ -330,11 +330,7 @@ fb_actions.remove_file = function(prompt_bufnr)
   print "\n"
 
   vim.ui.input({ prompt = "Remove selected files [y/N]: " }, function(input)
-    if input == nil then
-      return
-    end
-    answer = string.lower(input)
-    if answer == "y" then
+    if input:lower() == "y" then
       for _, p in ipairs(selections) do
         local is_dir = p:is_dir()
         p:rm { recursive = is_dir }
@@ -344,9 +340,11 @@ fb_actions.remove_file = function(prompt_bufnr)
         else
           fb_utils.delete_dir_buf(p:absolute())
         end
-        a.nvim_notify(string.format("\n%s has been removed!", p:absolute()), vim.log.levels.INFO, {})
+        print(string.format("\n%s has been removed!", p:absolute()))
       end
       current_picker:refresh(current_picker.finder)
+    else
+      print("Removing files aborted!")
     end
   end)
 end
--- a/lua/telescope/_extensions/file_browser/actions.lua
+++ b/lua/telescope/_extensions/file_browser/actions.lua
@@ -344,7 +344,7 @@ fb_actions.remove_file = function(prompt_bufnr)
       end
       current_picker:refresh(current_picker.finder)
     else
-      print("Removing files aborted!")
+      print(" Removing files aborted!")
     end
   end)
 end
--- a/lua/telescope/_extensions/file_browser/actions.lua
+++ b/lua/telescope/_extensions/file_browser/actions.lua
@@ -344,7 +344,7 @@ fb_actions.remove_file = function(prompt_bufnr)
       end
       current_picker:refresh(current_picker.finder)
     else
-      print(" Removing files aborted!")
+      print " Removing files aborted!"
     end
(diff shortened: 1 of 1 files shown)


lua/telescope/_extensions/file_browser/actions.lua
```diff
@@ -329,8 +329,12 @@ fb_actions.remove_file = function(prompt_bufnr)
   -- format printing adequately
   print "\n"
 
-  vim.ui.select({ "Yes", "No" }, { prompt = "Remove Selected Files" }, function(_, idx)
-    if idx == 1 then
+  vim.ui.input({ prompt = "Remove selected files (Y/N): " }, function(user_answer)
```
jamestrew, line 332: I commonly see the yes/no options shown with the capitalized letter indicating the default option so in this case something like `[y/N]`. Otherwise I like this change.

fdschmidt93: Slight nitpick but I agree.

lua/telescope/_extensions/file_browser/actions.lua
```diff
@@ -340,7 +344,7 @@ fb_actions.remove_file = function(prompt_bufnr)
         else
           fb_utils.delete_dir_buf(p:absolute())
         end
-        print(string.format("%s has been removed!", p:absolute()))
+        a.nvim_echo({ { string.format("\n%s has been removed!", p:absolute()) } }, false, {})
```
fdschmidt93, line 347: I'm leaning towards switch printing / echoing to `vim.notify`. That way, users can have nicer notifications if they want to, and as I understand it (I'm not using it just yet), it's otherwise if not overridden by eg [nvim-notify](https://github.com/rcarriga/nvim-notify) practically equivalent to `nvim_echo`?

See also `:h nvim_notify`.

Thoughts?

lua/telescope/_extensions/file_browser/actions.lua
fdschmidt93, line 332: Slight nit for conventions: can we rename `user_answer` to `input` to stay streamlined with `vim.ui.input` docs?

What about also providing`completion` with `y/N` and `default = y`. In case of the latter, users can then just confirm with `CR`.

Thoughts?

fdschmidt93 (commented):
A couple of nits/comments but otherwise LGTM, thanks :)

lua/telescope/_extensions/file_browser/actions.lua
miguelmoraperea: > Slight nit for conventions: can we rename user_answer to input to stay streamlined with vim.ui.input docs?
Sure, that is not a problem.

> What about also providing completion with y/N and default = y. In the case of the latter, users can then just confirm with CR.

If you want to default as yes, then it should be `(Y/n)`.

However, I don't agree with having **Yes** as the default, if this was a `continue` or `copy` operation, then **Yes** as default is ok, but for `delete` operations we should probably default to **No** for safety to prevent deletions by accident.

All other plugins handle default as **No** for delete operations, we should follow the same in my opinion.

lua/telescope/_extensions/file_browser/actions.lua
miguelmoraperea: I'm ok with using `nvim_notify`, but I would suggest updating all instances of `nvim_echo` in the repo to use `nvim_notify` in a single commit, and then use `nvim_notify` from that point on. To keep it consistent instead of having echo in some places and notify in others.

That way we could also avoid nitpicking this API in future PRs. Any developers working on the project can see that we use nvim_notify and they would use it as well (hopefully, fingers crossed), I think that way we could be more efficient.

fdschmidt93: Agreed.

fdschmidt93 (commented):
Apart from `notify` (my bad), LGTM thanks :)

I'd merge then.

fdschmidt93 (approved):
Thanks :)

Issue #43, opened 2021-12-31
Switch from file/folder mode changes current directory back to 'cwd'


miguelmoraperea:
### Description

Switch from file/folder mode changes current directory back to `cwd`, that seems a bit weird, I would expect the current directory to remain the same. Ultimately, as a user, I only want to change the mode from where I am, I don't want to go back to `cwd` unless I specifically ask for it.

### Neovim version

```markdown
NVIM v0.7.0-dev+792-gc46f7caad
Build type: RelWithDebInfo
```

### Operating system and version

Ubuntu 20.04

### Steps to reproduce

1. Open file browser in file mode
2. Navigate to a sub-directory
3. Change to folder mode
4. The current directory goes back to `cwd`

### Expected behavior

1. Open file browser in file mode
2. Navigate to a sub-directory
3. Change to folder mode
4. The current directory remains the same, only change the mode to folder mode.

### Actual behavior

![change-to-folder-mode-goes-back-to-cwd](https://user-images.githubusercontent.com/42681212/147838434-2481dc12-833d-416d-82c7-3fd2cb158496.gif)

### Minimal config

```Lua
vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvim/site]]
local package_root = '/tmp/nvim/site/pack'
local install_path = package_root .. '/packer/start/packer.nvim'
local function load_plugins()
  require('packer').startup {
    {
      'wbthomason/packer.nvim',
      {
        'nvim-telescope/telescope.nvim',
        requires = {
          'nvim-lua/plenary.nvim',
          'nvim-telescope/telescope-file-browser.nvim',
        },
      },
      -- ADD PLUGINS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
    },
    config = {
      package_root = package_root,
      compile_path = install_path .. '/plugin/packer_compiled.lua',
      display = { non_interactive = true },
    },
  }
end
_G.load_config = function()
  require('telescope').setup()
  require('telescope').load_extension('file_browser')
  -- ADD INIT.LUA SETTINGS THAT ARE _NECESSARY_ FOR REPRODUCING THE ISSUE
end
if vim.fn.isdirectory(install_path) == 0 then
  print("Installing Telescope and dependencies.")
  vim.fn.system { 'git', 'clone', '--depth=1', 'https://github.com/wbthomason/packer.nvim', install_path }
end
load_plugins()
[...]

PR #44, opened 2021-12-31
FIX - Switch between file/folder mode (#43)

miguelmoraperea:
Before this commit, switching between file and folder browser mode
makes the finder return to the 'cwd' directory.

This behavior seems a bit odd, with this commit, instead of returning to
'cwd', the finder will remain in the current directory after switching
modes, this way the user can switch from file/folder mode without having
to navigate to the same directory again.

If a user wants to go back to 'cwd' they can always use 'C-W' to do
that.

### DIFF
--- a/lua/telescope/_extensions/file_browser/actions.lua
+++ b/lua/telescope/_extensions/file_browser/actions.lua
@@ -430,10 +430,6 @@ fb_actions.toggle_browser = function(prompt_bufnr, opts)
     local new_title = finder.files and "File Browser" or "Folder Browser"
     current_picker.prompt_border:change_title(new_title)
   end
-  if current_picker.results_border then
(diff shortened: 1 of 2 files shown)


miguelmoraperea:
After the changes the behavior is like this:

fix-change-to-folder-mode-remains-in-current-path

fdschmidt93:
This behavior is very much intentional, since folder browsing from `cwd` to low-level folders by means of fuzzy finding is in most scenarios always fast and enables file operations swiftly across many layers of directories  (also upwards from `finder.path`), which is a key feature of this extension. On the contrary, this change would make such workflows a lot more complicated (cf. gif on `README`, `<C-W>` I suppose now every time in between?) while seemingly offering hardly any practical gain. Or am I missing something?

What I agree though is that there should be a user-friendly mechanism to set `finder.cwd`, very much in line with #40 and/or enabling this behavior optionally.

fdschmidt93:
Superseded by #57.

PR #45, opened 2021-12-31
fix: Incorrect file copied message (#37)

miguelmoraperea:
When copying a file outside the original directory (cwd) the file is not
copied, this is expected, but the user still gets the following message:

  [file] has been copied!

This is incorrect, this commit fixes the issue by verifying the result
status from the copy operation and prints the appropriate message
'[file] has been copied!' or '[file] has not been copied' depending on
whether the operation succeeded or not.

### DIFF
--- a/lua/telescope/_extensions/file_browser/actions.lua
+++ b/lua/telescope/_extensions/file_browser/actions.lua
@@ -295,12 +295,23 @@ fb_actions.copy_file = function(prompt_bufnr)
       end
     end
     if destination ~= "" then -- vim.fn.input may return "" on cancellation
-      file:copy {
+      local res = file:copy {
         destination = destination,
         recursive = true,
         parents = true,
       }
-      print(string.format("\n%s has been copied!", filename))
+
+      for key, value in pairs(res) do
+        if key.filename == destination then
+          local res_msg
+          if res[key] then
+            res_msg = string.format("\n%s has been copied!", filename)
+          else
+            res_msg = string.format("\n%s has not been copied", filename)
+          end
+          a.nvim_echo({ { res_msg } }, false, {})
+        end
+      end
     end
   end
 


fdschmidt93:
As a heads up, I'll try to look at this by mid next week :) It very much looks like a good change but also not urgent and I just want to have one clean pass whether typical scenarios are properly handled.

lua/telescope/_extensions/file_browser/actions.lua
```diff
@@ -295,12 +295,23 @@ fb_actions.copy_file = function(prompt_bufnr)
...
         destination = destination,
         recursive = true,
         parents = true,
       }
-      print(string.format("\n%s has been copied!", filename))
+
+      for key, value in pairs(res) do
+        if key.filename == destination then
+          local res_msg
+          if res[key] then
```
fdschmidt93, line 307: One thing that comes to mind here (referring to my other post), for instance are copying folders.

IIRC my plenary PR correctly, a folder would return a table of 

```lua
{
  file_a.ext = bool,
  file_b.ext = bool,
  folder = { ... },
}
```

Hence, I think we'd be erroneously printing success, whenever one of the sub-files is copied improperly.

fdschmidt93:
Some broader thoughts to (I will also) have think more about:

* Should we rather print/notify failures rather than successes
* (More broadly), should we use `vim.notify` as opposed to printing/echoing?

This goes a bit into the thought of making current multi-selections more visible, see discussion as per #41

lua/telescope/_extensions/file_browser/actions.lua
miguelmoraperea: Thanks for pointing this out @fdschmidt93 

I think you are onto something, but actually, we are not printing any message for copied folders.

I also noticed something interesting when copying files/folders, we are copying only the files/folders in the current dir and the first sub-dir down, all other deeper levels are recursively copied over, but we don't get a result for those file/directories after the first level down.

I need to take a closer look into this. I'll try to work on this sometime tomorrow.

PR #47, opened 2022-01-05
fix: path and title display

ikerurda:
When path_display is changed in telescope's default setup, (in the following image to "smart"), file browser is also affected, which I don't really know if it makes any sense...

Actual behavior:
Screenshot 2022-01-05 at 10 39 11 PM

Desired (?) behavior:
Screenshot 2022-01-05 at 10 39 26 PM

(path_display is set to truncate = 3 because otherwise when in folder-browser, we cannot see the end of the path correctly)


lua/telescope/_extensions/file_browser/make_entry.lua
```diff
@@ -12,6 +12,7 @@ local make_entry = function(opts)
     local hl_group
     -- mt.cwd can change due to caching and traversal
     opts.cwd = mt.cwd
+    opts.path_display = { ["truncate"] = 3 }
```
fdschmidt93, line 15: How would the user now set `path_display` if they for some reason would want to have it otherwise? One reason here I think what's worth thinking about is `path_display` in the context of the `folder_browser` and whether we can improve there.

fdschmidt93 (commented):
I can reproduce the problem this PR is trying to solve. However, I can just as well fix it locally by 

```lua
require "telescope".setup {
  defaults = {
    path_display = { "smart" },
  },
  extensions = {
    file_browser = {
      path_display = { truncate = 3 },
      },
    },
  }
}
```

which is OK as an intermediate solution.

I'd strive for something more cleaner solution that also makes `path_display` worthwhile for `folder_browser`. So, in conclusion:

* Agreed, the outcome for `file_browser` would be the desired effect yes, but I don't want a band-aid fix since the above seems to work
* We need a solution that also plays well with `folder_browser`

In super quick test `path_display` doesn't seem nicely and unforunately, `make_entry` in `telescope-file-browser` is rather brittle (caching entries for multi selections, display, `path_display`, ...).

You're very welcome to take a crack :) If not, I'll try to look some more into this, though as a heads up I'll be travelling and largely unavailable for 3 weeks.

ikerurda:
Ok, I agree. Don't know where to start but I'll try to take a look soon. Tranks!

ikerurda:
I've made "tails" the default for path_display in file_browser and allowed user configs in folder_browser again.

Still, with "smart" path display in folder_browser, there seems to be something odd ("/telescope-file-browser.nvim" is common in every entry and therefore shouldn't be showing):
Screenshot 2022-01-07 at 9 48 11 PM

Could this be because we are using `require"telescope.utils".transform_path(...)` incorrectly in `make_entry.lua` or is it because of an upstream error?

Also, in folder_browser mode, the title seems too long. I'm thinking of applying there also user's path_display. What do you think?

fdschmidt93:
Just skimming. Are the changes related to titles fix ups for #40? That should go into a separate PR.

I'm ok merging the `tail` for `file_browser` as a band-aid with a TODO comment. I'll have to refactor the entry maker stuff eventually in larger scope. As mentioned, it's quite brittle.

ikerurda:
Not just fixing #40 but other actions that weren't updating the titles correctly, and I'm also truncating long titles.
I can move it to a new PR if you want, I just thought this may be related...

Apart from that, do you have any idea why this is happening? Is this because of an upstream issue?
> Still, with "smart" path display in folder_browser, there seems to be something odd ("/telescope-file-browser.nvim" is common in every entry and therefore shouldn't be showing): Screenshot 2022-01-07 at 9 48 11 PM

fdschmidt93:
Just a heads up, I plan on looking into this next week. Probably enough file browser for me this week, it was quite some work tonight :sweat_smile:

lua/telescope/_extensions/file_browser/utils.lua
```diff
@@ -118,6 +119,8 @@ fb_utils.redraw_border_title = function(current_picker)
     else
       new_title = finder.cwd
     end
+    local width = vim.api.nvim_win_get_width(0) * 0.80
```
fdschmidt93, line 122: Slight nit. `0` here refers to the prompt window rather than the results window. In, for instance, `ivy` theme, they are of different width. Not sure if they otherwise can differ but let's refer to the correct window just in case like

```lua
local width = math.floor(a.nvim_win_get_width(current_picker.results_win) * 0.8)
```

should do the job as intended.

fdschmidt93 (commented):
LGTM thanks :)

For the future, please only one thing at a time. It's fine here since I'll try to clean up `make_entry` soon.

ikerurda:
Sorry, this is all pretty new to me... :)

fdschmidt93 (approved):
Thanks :)

PR #48, opened 2022-01-07
feat: toggle view of multi selections

fdschmidt93:
cc @jamestrew you're happy to give this a whirl and give feedback

https://user-images.githubusercontent.com/39233597/148565206-e9ae5dea-cf9a-4d16-92da-d55070df23d5.mp4

### DIFF
--- a/doc/telescope-file-browser.txt
+++ b/doc/telescope-file-browser.txt
@@ -265,6 +265,17 @@ fb_actions.select_all({prompt_bufnr})                *fb_actions.select_all()*
         {prompt_bufnr} (number)  The prompt bufnr
 
 
+fb_actions.toggle_selections({prompt_bufnr})  *fb_actions.toggle_selections()*
+    Toggle between view of multi-selections and previous browser.
+    - Note:
+      - WARNING: many |fb_actions| might not work.
+      - De-selections are persisted to other finders!
+
+
+    Parameters: ~
+        {prompt_bufnr} (number)  The prompt bufnr
+
+
 
 ================================================================================
                                                 *telescope-file-browser.finders*
--- a/lua/telescope/_extensions/file_browser.lua
+++ b/lua/telescope/_extensions/file_browser.lua
@@ -73,6 +73,7 @@ local pconf = {
       ["<C-s>"] = fb_actions.goto_home_dir,
       ["<C-t>"] = fb_actions.change_cwd,
       ["<C-y>"] = fb_actions.copy_file,
+      ["<C-k>"] = fb_actions.toggle_selections,
     },
     ["n"] = {
       ["dd"] = fb_actions.remove_file,
@@ -87,6 +88,7 @@ local pconf = {
       ["s"] = fb_actions.goto_home_dir,
       ["t"] = fb_actions.change_cwd,
       ["y"] = fb_actions.copy_file,
+      ["k"] = fb_actions.toggle_selections,
     },
   },
   attach_mappings = function(prompt_bufnr, _)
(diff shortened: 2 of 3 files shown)


Conni2461:
I dont know if this is a good for core (at least right now) but i think its cool and i know how we could improve it.

So its similar to this thing https://github.com/nvim-telescope/telescope.nvim/issues/1590  but rather than piping in all results we pipe in all multi_selection (basically just a subset).

So what we could do, to reduce code duplication and to make this more awesome :laughing: Is something like this

```lua
require("telescope.actions.generate").pipe(prompt_bufnr, {
  prompt_title = "Stuff" -- optional
  results_title = "More stuff"
  sorter = conf.generic_sorter({}), -- optional
  finder = finders.new_iterator_finder { 
    results = ipairs(current_picker:get_multi_selection()),
    entry_maker = ... 
  },
  -- or. i think entry manager doesn't have a ipairs iterator (yet)
  finder = finders.new_iterator_finder { 
    results = ipairs(current_picker.manager.ipairs()), 
    entry_maker = ... 
  },
})
```

So why am I thinking iterator, because right now you need to first iterate over all items put them into a table so the finder can iterate over them again to put them into the next table. 

So having this more high level interface (of refresh) would clean up this PR and it should still be possible to do the toggling thing. Also it would be easier to maintain for the future.

Thoughts

jamestrew:
It looks good but like you mentioned, it does cause all sorts of strange behaviors with I think most of the actions. 

* `toggle_all`: seems fine (occasionally glitch-y graphically)
* `remove_file`: good
* `create_file`:  error - `finder.path` is nil
* `toggle_browser`:  no error but still broken - again probably related to `finder.path` is nil
* `goto_parent_dir`:  nothing happens
* `goto_parent_dir`: 
* `goto_parent_dir`: 
* `goto_parent_dir`: 
*

jamestrew:
It looks good but like you mentioned, it does cause all sorts of strange behaviors with a good chunk of the actions. 

Documenting the action behaviors here:
* `toggle_all`: seems fine (occasionally glitch-y graphically which we already briefly discussed)
* `remove_file`: functioning as intended
* `create_file`: error - `finder.path` is nil
* `toggle_browser`: no error but still broken - again probably related to `finder.path` is nil
* `goto_parent_dir`: nothing happens
* `toggle_hidden`: nothing happens
* `open_file`: with nothing selected, gives helpful message. With things selected, closes Telescope
* `rename_file`: functioning as intended
* `goto_cwd`: changes `results_title` to `./`
* `goto_home_dir`: changes `results_title` to `nil`
* `change_cwd`: error with no selection
* `copy_file`: pretends to work "... has been copied!" but obviously not the intended usage 
* `toggle_selections`: functioning as intended

I think many of these don't really have to be functional during the `Multi Selection` view. I think the actions which obviously should work are `toggle_all`, `remove_file`, `toggle_browser`, `rename_file` and `toggle_selections`. Maybe a couple others could be included. The rest, we can maybe get away with a message like `[telescope] Return to file/folder browser to perform desired action` and suppress/exit the action?

fdschmidt93:
@Conni2461 fully agree. Sounds like that would amount to quite the PR though :sweat_smile: I'll try to have a look into this more next week

@jamestrew thanks for the bunch of tests! I'd say this would be a rather tedious way to deal with this. Such use cases kind of make me believe maybe mappings should be associated with finders as opposed to prompt buffers and then you'd just deactivate all of the `fb_actions`. I'll see whether there is a holistic way to deal with these when I get around to doing the telescope PR I suppose.

PR #52, opened 2022-01-11
chore: overhaul documentation

fdschmidt93:
~~Sorry folks, this includes a breaking change to remove trailing `_file` of corresponding `fb_actions` that comprise it to avoid future confusion as maybe some user might think actions are dedicated to files.~~

Will follow later tonight


fdschmidt93:
Fingers crossed all is well and this just represents a massive improvement :)

PR #49, opened 2022-01-08
rft!: mappings (suggestion)

ikerurda:
I propose using `<A-` for file management (add, rename, move, yank, delete)
and `<C-` for general file/folder-browsing.

| Insert / Normal | Action                                                           |
|-----------------|------------------------------------------------------------------|
| `<A-c>/c`       | Create file/folder at current path (trailing `/` creates folder) |
| `<A-r>/r`       | Rename multi-selected files/folders                              |
| `<A-m>/m`       | Move multi-selected files/folders to current path                |
| `<A-y>/y`       | Copy multi-selected files/folders to current path                |
| `<A-d>/d`       | Delete multi-selected files/folders                              |
| `<C-o>/o`       | Open file/folder with default system application                 |
| `<C-g>/g`       | Go to parent directory                                           |
| `<C-e>/e`       | Go to home directory                                             |
| `<C-w>/w`       | Go to current working directory (cwd)                            |
| `<C-t>/t`       | Change nvim's cwd to selected folder/file(parent)                |
| `<C-f>/f`       | Toggle between file and folder browser                           |
| `<C-h>/h`       | Toggle hidden files/folders                                      |
| `<C-a>/a`       | Toggle all entires ignoring `./` and `../`                       |

Thoughts?

PR #56, opened 2022-01-11
break!: revamp mappings

fdschmidt93:
Continuing #49 with the slight change of toggling to `s` to not affect changing to insert mode, that one would bite me too often


fdschmidt93:
Already apologies to anyone who doesn't like this. Now there is some more logic to default mappings, which are primarily intended for novice and new users to simplify discoverability, as I trust experienced users can easily remap stuff within 5 minutes. If enough users speak up I'm happy to change this again though. I personally don't care so much, but prefer that we now have some logic in place.

In any case, remapping should be very straightforward.

fdschmidt93:
Please direct any creds (or blame :laughing: ) to @ikerurda 

Thanks :)

PR #62, opened 2022-01-12
fix!: goto parent dir default key to <C-b>/b

fdschmidt93:
As per reddit comment.

I'm leaning towards `<C-b>/b` as that allows for the insert/normal mode mirroring of other keymaps rather than `-` (e.g. `nvim-tree.lua` default), since `<C-->` is not a recognized keycode.

Please hit an emoji on this PR with

| Emoji | Insert/normal mode mapping|
|-------|---------------------------|
| Rocket| `<C-b>/b`                 |
| Celebrate | `-/-` |

while keeping mapping convention at least somewhat in mind and that classic file tree's don't have to distinguish between insert and normal mode all too much I guess :)

For the time being I personally favor `b`.

Rules to break my decision:

* 10+ participants
* ~60% in favor of `-`

I'll let this run I guess for a couple of days.

Thanks!


tjdevries:
I use `-` :)

ChadCapra:
I use tmux default bindings so <C-b> prefix will conflict for me, but thank you for the awesome plugin!

PR #68, opened 2022-01-15
feat: group dirs and files in file_browser

fdschmidt93:
As per reddit request. This PR introduces the `grouped` option which initially groups entries in the `file_browser` by directory and files and only then sorts alphabetically. I would probably mark this optional as this is behavior that cannot be easily achieved with an `fd` finder, but happy to be convinced otherwise.

`fd` is required for `depth=false` and fast folder browsing. For classic `file_browser`, `plenary.scandir` is already plenty fast.

Before:

before

After

after

### DIFF
--- a/doc/telescope-file-browser.txt
+++ b/doc/telescope-file-browser.txt
@@ -83,6 +83,9 @@ fb_picker.file_browser({opts})                      *fb_picker.file_browser()*
         {cwd_to_path}       (boolean)   whether folder browser is launched
                                         from `path` rather than `cwd`
                                         (default: false)
+        {grouped}           (boolean)   group initial sorting by directories
+                                        and then files; uses plenary.scandir
+                                        (default: false)
         {files}             (boolean)   start in file (true) or folder (false)
                                         browser (default: true)
         {add_dirs}          (boolean)   whether the file browser shows folders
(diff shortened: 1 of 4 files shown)


prdanelli:
I seem to be getting the sorting in the "preview" pane, but not in the "fuzzy finder" section on the left of the screenshot.

image

Am I doing something wrong? I'm adding this as my setup: 

```
telescope.load_extension("file_browser", { grouped = true })
```
And calling the extension with the following: 
```
vim.api.nvim_set_keymap("n", "<leader>ff", ":Telescope file_browser<cr>", { noremap = true })
```

ikerurda:
Hi!
You can set it by defaul:
```lua
require"telescope".setup {
  ...
  extensions = {
    file_browser = {
      ...
      files = false,
    }
  }
}
```

Via the vim command:
`:Telescope file_browser grouped=true`

Or with the lua API:
`:lua require"telescope".extensions.file_browser.file_browser( { grouped = true } )`

Read the [readme](https://github.com/nvim-telescope/telescope-file-browser.nvim#setup-and-configuration)section  for more info :)

prdanelli:
Thank you @ikerurda, the key maps worked great.

PR #72, opened 2022-01-17
refactor: standardize fb_action user IO

jamestrew:
Using `vim.notify` for output messages (log level = WARN for areas indicating missteps by users), `vim.ui.input` for inputs.

All messages prefixed by `[telescope]` to be clear about the origins of the message.

Checks off an item from #3

### DIFF
--- a/lua/telescope/_extensions/file_browser/utils.lua
+++ b/lua/telescope/_extensions/file_browser/utils.lua
@@ -147,4 +147,46 @@ fb_utils.group_by_type = function(tbl)
   end)
 end
 
+local function add_tele_prefix(msg)
+  if not msg:match "^%[telescope%]" then
+    msg = "[telescope] " .. msg
+  end
+  return msg
+end
+
+-- Move this and above functions into telescope core?
+fb_utils.tele_notify = function(msg, log_level, opts)
+  while msg:match "^\n" do
+    msg = msg:sub(2, #msg)
+    vim.notify "\n"
+  end
+  vim.notify(add_tele_prefix(msg), log_level, opts)
+  while msg:match "\n$" do
+    msg = msg:sub(1, #msg - 1)
+    vim.notify "\n"
+  end
+end
+
+fb_utils.get_valid_path = function(prompt, default, check_exist)
+  check_exist = vim.F.if_nil(check_exist, true)
+  local path
+  vim.ui.input({ prompt = add_tele_prefix(prompt), default = default }, function(input)
+    input = input and input:match "^%s*(.-)%s*$" or nil
+    if not input then
+      fb_utils.tele_notify("\nOperation aborted!", vim.log.levels.WARN)
+      return
+    elseif input == "" then
+      fb_utils.tele_notify("\nPlease provide valid path input!", vim.log.levels.WARN)
+      return
+    end
+
+    path = Path:new(input)
+    if check_exist and path:exists() then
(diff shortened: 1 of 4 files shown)


lua/telescope/_extensions/file_browser/utils.lua
```diff
@@ -147,4 +147,46 @@ fb_utils.group_by_type = function(tbl)
...
 end
 
+local function add_tele_prefix(msg)
+  if not msg:match "^%[telescope%]" then
+    msg = "[telescope] " .. msg
+  end
+  return msg
+end
+
+-- Move this and above functions into telescope core?
```
tami5, line 157: Makes sense

tami5 (commented):
awesome refactor

fdschmidt93:
Sorry, a bit too lazy right now as I've had worked on tons on the file browser in the weeks before. I'll try to get to have a first brief crack at this either tomorrow or Sunday :)

lua/telescope/_extensions/file_browser/utils.lua
```diff
@@ -147,4 +147,46 @@ fb_utils.group_by_type = function(tbl)
...
+fb_utils.tele_notify = function(msg, log_level, opts)
```
Conni2461, line 158: I dont get what you try to achieve with both of these functions. Any styling should be handled by vim.notify, thats the whole point.

For example i saw this:
```
fb_utils.tele_notify "\nRemoving files aborted!"
```

Which results in this output. If you are using https://github.com/rcarriga/nvim-notify

image

the max wrapper around vim.notify i would accept in telescope core is this

```lua
utils.notify = function(msg, log_level, opts)
  vim.notify("[telescope]" .. msg, log_level, opts)
end
```

And then we could replace all the print 
```lua
print "[telescope] Nothing currently selected"
```
with 
```lua
utils.notify "Nothing currently selected"
```
in core

Conni2461: Or we can use the title possibility of `nvim-notify` if its available. So if not `[telescope] msg` else `vim.notify(msg, log_level, { title = "telescope" })`

tami5: > Or we can use the title possibility of `nvim-notify` if its available. So if not `[telescope] msg` else `vim.notify(msg, log_level, { title = "telescope" })`

😋🎉  bingo

jamestrew:
Throwing this PR into draft mode to do some clean up and find a solution to #87, #91

jamestrew:
closes #91 

I've spent some more time looking for solutions for #87 but I'm able to find something that works consistently.
So far, I'm using `vim.api.nvim_command("norm :esc<CR>")` and it works for all actions expect `remove` (due to things being printed prior to the `[y/N]` input getter).  Also for `get_valid_path` if the user completely deletes the default input, it will also print the `Operation aborted` msg on the same line as the input prompt.

fdschmidt93:
I've now looked a bit more at this as I'll eventually will take this over. Couple of notes:

- `vim.ui.input` is async for most users that eg user `dressing.nvim` or `telescope-ui-select`; hence, we cannot rely returning anything from `vim.ui.input`
- `vim.cmd [[redraw!]]` is probably the most reliant way to avoid `hit-enter` prompt

I'll try to work incrementally and push the PR maybe in a week or two.

Thanks again anyways :)

PR #73, opened 2022-01-17
feat: folder_browser with more filetypes

fdschmidt93:
This allows the following:

`:lua require "telescope".extensions.file_browser.file_browser { types = {"file", "directory"}}`

now makes the `folder_browser` show both files and directories (default: `types = { "directory" }`) and you can switch to the folder of a particular file.

https://user-images.githubusercontent.com/39233597/149786928-82d8d0af-2704-4c9e-be60-60c24961097f.mp4


windowsrefund:
This seems like it would be a great solution for #176.

PR #74, opened 2022-01-17
feat: terminal cmd preview for directories

fdschmidt93:
Ah, this is a tricky one. This allows overwriting the directory previewer with a function like so

```lua
function(path)
  return "ls -la --color=always " .. path .. " | less -r"
end
```

would show `ls -la` output like in the terminal

The video shows `tree` as a previewer. Colors may or may not work depending on compatability with a pager.

https://user-images.githubusercontent.com/39233597/149843316-89d4e917-aefe-4fa0-bb50-94cc9e401730.mp4

I hope someone appreciates this hack :laughing:

### DIFF
--- a/lua/telescope/_extensions/file_browser/picker.lua
+++ b/lua/telescope/_extensions/file_browser/picker.lua
@@ -18,7 +18,7 @@ local pickers = require "telescope.pickers"
 local conf = require("telescope.config").values
 
 local fb_finder = require "telescope._extensions.file_browser.finders"
-
+local fb_previewer = require "telescope._extensions.file_browser.previewer"
 local Path = require "plenary.path"
 local os_sep = Path.path.sep
 
@@ -67,7 +67,7 @@ fb_picker.file_browser = function(opts)
     prompt_title = opts.files and "File Browser" or "Folder Browser",
     results_title = opts.files and Path:new(opts.path):make_relative(cwd) .. os_sep or "Results",
     finder = fb_finder.finder(opts),
-    previewer = conf.file_previewer(opts),
+    previewer = fb_previewer.previewer.new(opts),
     sorter = conf.file_sorter(opts),
   }):find()
 end
--- a/lua/telescope/_extensions/file_browser/previewer.lua
+++ b/lua/telescope/_extensions/file_browser/previewer.lua
@@ -1,0 +1,77 @@
+local from_entry = require "telescope.from_entry"
+local Path = require "plenary.path"
+local conf = require("telescope.config").values
+
+local fb_previewer = {}
+local buffer_previewer = require("telescope.previewers.buffer_previewer").cat
+local term_previewer = require "telescope.previewers.term_previewer"
+
+local action_state = require "telescope.actions.state"
+
+local function defaulter(f, default_opts)
+  default_opts = default_opts or {}
+  return {
+    new = function(opts)
+      if conf.preview == false and not opts.preview then
+        return false
+      end
+      opts.preview = type(opts.preview) ~= "table" and {} or opts.preview
+      if type(conf.preview) == "table" then
+        for k, v in pairs(conf.preview) do
+          opts.preview[k] = vim.F.if_nil(opts.preview[k], v)
+        end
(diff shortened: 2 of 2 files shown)


ikerurda:
Cool!
But when quitting the picker in insert mode, it stays in insert mode... not sure why

fdschmidt93:
Thanks for test driving!

I can reproduce that, but not really sure as to why that would happen :sweat_smile:

In general, it seems to be associated with terminal previewers, e.g.

`:lua require "telescope.builtin".find_files { previewer = require "telescope.previewers.term_previewer".cat.new({})}` also has that issue.

Probably related to the prompt buffer bugs.

fdschmidt93:
The fix is quite hacky, but seems to work as intended so long as anyone would want to return to insert mode coming off of the `file_browser`. I'd probably let anyone who ever (if ever anyone) runs into the issue to fix that.

ikerurda:
Yes, this fixes it :)

tami5:
😍This just what I was looking for. does it need more tests?

fdschmidt93:
No, primarily documentation so people will understand what they can pass as `dir_preview` then.

I'll try to do that tonight.

Conni2461:
We should drop scandir.ls and everything with it. It was a mistake. Makes highlighting for unix like and windows systems the same but offers 0 flexibility. Windows just only gives us `dirs`

But `new_termopen_previewer` is still leaking buffers. Just do `:ls!`
A solution would be `new_buffer_previewer` with `vim.api.nvim_open_term` as a new `term_previewer`. Would also remove the `ls -r` part.

I would suggest to not do this here but rather in core

tami5:
> We should drop scandir.ls and everything with it. It was a mistake. Makes highlighting for unix like and windows systems the same but offers 0 flexibility. Windows just only gives us `dirs`
> 
> But `new_termopen_previewer` is still leaking buffers. Just do `:ls!` A solution would be `new_buffer_previewer` with `vim.api.nvim_open_term` as a new `term_previewer`. Would also remove the `less -r` part.
> 
> I would suggest to not do this here but rather in core

No way it should be dropped @Conni2461  it extremely fast and we have end up creating it for a reason. Yah I agree on the flexibility part but that just a render function need to be exposed and a set of themes to be built upon it. It's extremely easy t build tree structure or hide certain information. 

I can take a look on increasing it flexibility if you like.

fdschmidt93:
One thing I've planned on doing eventually for `telescope-file-browser.nvim` is expose the underlying methods of `plenary.scandir.ls` such that the `make_entry` of the `file_browser` can make use of those.

The idea is that `make_entry` into some kind of builder pattern such that users can customize it more to their liking and possible toggle, for instance.

1. Open file browser
2. Toggle file size
3. Be able to sort by file size
4. Toggle off file size

What I nevertheless really like would be incorporating `nvim_open_term` into the buffer previewer. We could provide a hook (they are in need of a clean up anyways I suppose, should probably be `hook = { filetype = function(...) end, filesize = function(...) end` rather than individual functions), for conditional terminal preview. I think that way, we can maybe fully deprecate term-previewers from telescope and just have one properly clean main previewer interface. Unifies scrolling, cleanup, etc. But maybe I'm missing something.

fdschmidt93:
I'll be working on this upstream most likely in a more universal solution. Let's close this.

yingzhu146:
@fdschmidt93 this looks amazing! I think this (tree as command) is also my interpretation of what #165 was hinting at which I'd love since that's the default behavior of https://github.com/Canop/broot which I currently still use for exploration.

Do you think this could be useful for non-directories too?

https://github.com/jarun/nnn allows specifying it per file-type via plugin https://github.com/jarun/nnn/blob/master/plugins/preview-tui

I think that would be extremely powerful. Just some ideas for use cases


directory preview:

tree via `exa --tree --icons`

file preview:

`head -l50 | hexyl`: binaries
`vd` for csv (https://www.visidata.org/)
`fx` for json 
`euporie --dump` for jupyter notebooks (https://github.com/joouha/euporie)

PR #81, opened 2022-01-21
Feat: Bookmarks

ikerurda:
What do you think of this feature?
Would allow users to set some bookmarks:
      
```lua
require("telescope").setup {
  ...
  extensions = {
    file_browser = {
      bookmarks = {
        ["display name"] = "~/some/path",
        ["config"] = "~/.config/nvim",
      },
    },
  },
}
```

Users can currently select a bookmark with the mapping `<c-b>/b`.
This action only changes the path of the finder, not the cwd (this could be configurable in the future).

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,9 +45,11 @@ require("telescope").setup {
           -- your custom normal mode mappings
         },
       },
-      bookmarks = {
-        ["display name"] = "~/some/path",
-        ["config"] = "~/.config/nvim",
+      user_data = {
+        bookmarks = {
+          ["display name"] = "~/some/path",
+          ["config"] = "~/.config/nvim",
+        }
       }
     },
   },
--- a/README.md
+++ b/README.md
@@ -46,10 +46,7 @@ require("telescope").setup {
         },
       },
       user_data = {
-        bookmarks = {
-          ["display name"] = "~/some/path",
-          ["config"] = "~/.config/nvim",
-        }
+        -- data you may want to store for your custom actions (access from finder.user_data)
       }
     },
   },
--- a/README.md
+++ b/README.md
@@ -45,6 +45,10 @@ require("telescope").setup {
           -- your custom normal mode mappings
         },
       },
+      bookmarks = {
+        ["display name"] = "~/some/path",
+        ["config"] = "~/.config/nvim",
+      }
     },
   },
 }
--- a/doc/telescope-file-browser.txt
+++ b/doc/telescope-file-browser.txt
@@ -86,8 +86,6 @@ fb_picker.file_browser({opts})                      *fb_picker.file_browser()*
         {grouped}           (boolean)   group initial sorting by directories
                                         and then files; uses plenary.scandir
                                         (default: false)
-        {bookmarks}         (table)     bookmarks to select from (default:
-                                        empty)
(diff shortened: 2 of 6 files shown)


fdschmidt93:
I think it's cool, but apologies, I don't think this is something that needs to be built into `telescope-file-browser.nvim` as it's much more general and theoretically could apply to any telescope picker for that matter.

So I'd rather see this spun into a separate extension library, and then I suppose you arrive at something very similar to `telescope-project.nvim`.

In other words, I'm afraid the only value added here maybe would maybe be having `telescope-file-browser.nvim` bookmarks, specifically.

What I may see as useful directly relating to `telescope-file-browser.nvim` is the ability to register keys to the finder, to e.g. store bookmarks or meta-data the user may need for their own actions. I'm not sure though whether `telescope.nvim` itself has something akin to that already.

ikerurda:
On the one hand, I agree, it is very similar to `telescope-project.nvim`. But I was thinking that we could offer this functionality as it is just an action and would prevent users from having to install _another_ extension.
Still, I see your point and I'm ok with it.

On the other hand, I don't get your point here:

> ... having `telescope-file-browser.nvim` bookmarks, specifically.
> 
> What I may see as useful directly relating to `telescope-file-browser.nvim` is the ability to register keys to the finder, to e.g. store bookmarks or meta-data the user may need for their own actions.

fdschmidt93:
If you can pass a table of data to the finder or picker, anyone wanting this feature could quite easily with a bit of knowledge of lua and telescope(filde-browser) internals, whip up such a bookmark feature.

ikerurda:
Is this what you meant?

Users could then use something like
```lua
local open_bookmark = function(prompt_bufnr)
  local fb_utils = require "telescope._extensions.file_browser.utils"
  local bookmarks = finder.user_data.bookmarks

  if not bookmarks or vim.tbl_isempty(bookmarks) then
    print "[telescope] Please set some bookmarks first"
    return
  end

  vim.ui.select(
    vim.tbl_keys(bookmarks),
    { prompt = "Select bookmark:" },
    function(selection)
      if selection then
        finder.path = vim.fn.expand(bookmarks[selection])
        fb_utils.redraw_border_title(current_picker)
        current_picker:refresh( finder, { reset_prompt = true, multi = current_picker._multi })
      end
    end
  )
end
```

Would you add this to the wiki?

fdschmidt93:
> Would you add this to the wiki?

Sure, you should be able to add this to the wiki yourself. Anyone can edit it. If that's not the case, let me know.

Apologies, I've also thought about this once more. While I believe that something like `user_data` may be interesting, in this case, it's actually not required.

`bookmarks` could just be any table in the users configuration. No need for the `finder` or `picker` to carry that around. Nevertheless, on probably picker dependent attributes, it makes sense I guess, though not that I can think of any. In any case, that would be a `telescope` PR I suppose. And even there we should have demonstrate a clear use case for that.

I'm very sure that anyone desiring bookmarks should probably use `telescope-project` (which, IIRC, even has a builtin action to launch the (deprecated) file browser in a project folder).

Apologies if that's a rather frustrating outcome. The value-add is just too slim. Ideally, we can chat about such proposals in feature requests beforehand so you don't have to jump through extra hoops.

PR #94, opened 2022-01-31
feat: add option `hide_parent_dir`

### LINKED ISSUE #93: add option to hide parent entry `..`
Since there is the action `goto_parent_dir`, in most cases, `..` entry is unnecessary.

jamestrew:
Adds new option `hide_parent_dir` (default false) to hide `../` in the file browser as it can be redundant to experienced users who use the `goto_parent_dir` action.

closes #93

### DIFF
--- a/lua/telescope/_extensions/file_browser/picker.lua
+++ b/lua/telescope/_extensions/file_browser/picker.lua
@@ -54,6 +54,7 @@ local fb_picker = {}
 ---@field respect_gitignore boolean: induces slow-down w/ plenary finder (default: false, true if `fd` available)
 ---@field browse_files function: custom override for the file browser (default: |fb_finders.browse_files|)
 ---@field browse_folders function: custom override for the folder browser (default: |fb_finders.browse_folders|)
+---@field hide_parent_entry boolean: hide `../` in the file browser (default: false)
 fb_picker.file_browser = function(opts)
   opts = opts or {}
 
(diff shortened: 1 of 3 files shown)


lua/telescope/_extensions/file_browser/picker.lua
```diff
@@ -54,6 +54,7 @@ local fb_picker = {}
 ---@field respect_gitignore boolean: induces slow-down w/ plenary finder (default: false, true if `fd` available)
 ---@field browse_files function: custom override for the file browser (default: |fb_finders.browse_files|)
 ---@field browse_folders function: custom override for the folder browser (default: |fb_finders.browse_folders|)
+---@field hide_parent_entry boolean: hide `../` in the file browser (default: false)
```
fdschmidt93, line 57: Let's call this `hide_parent_dir`.

It's a bit clearer as not everyone is familiar with telescope terminology.

fdschmidt93 (commented):
Thanks for the PR :)

Rather than not returning a result in the entry maker, this should be done here, maybe with slight refactoring to make it cleaner.

https://github.com/nvim-telescope/telescope-file-browser.nvim/blob/e65a5674986dcaf27c0fd852b73f5fc66fa78181/lua/telescope/_extensions/file_browser/finders.lua#L29-L71

The reasoning is:

- While checking is fast, we now potentially check thousands of times, possibly even unrelated (`folder_browser`)
- Null result skews counting, ie not perfectly clean
- It's really just boils down to not creating the entry in the finder to begin with

fdschmidt93 (approved):
LGTM thanks :)

PR #96, opened 2022-02-01
close #95: add option `select_buffer`

### LINKED ISSUE #95: add option to auto focus on the entry of current buffer when the displayed directory is the directory of current file
## Reasons

A **new operation mode** is entered when fb is started with its own key mappings, it is glad for me to map '-' to open fb **in the  current directory** and use the buffer key mappings. For example, I set the initial mode to `normal` and map `L` to open on the right, and it will **not occupy** the mapping of `<m-l>` or `<c-l>` . It is more convenient than run commands in the command line! But I think there can be a little improvement, it will be better if it can try to focus on the current buffer when someone prefers to **operate on the file of current buffer**.

Since fb can make use of the **fuzzy search** of upstream, the initial selection of entry may be not so important.

## Some details

The possible way is to provide an option such as `auto_focus`, so that users can set it to true in the `setup` function or **pass it to the startup function**.

Note that hidden files need to be toggled if the file of current buffer is a hidden file.

Well, it is better to also control `auto_focus` by adding a new action `auto_focus_toggle` :)

ZhiyuanLck:
If the option is set to `true`, file browser will try to select the
current buffer at start. It has no effect if `opts.path` or `opts.cwd`
is not the parent directory of the current buffer or `opts.depth` is
greater than 1 or the initial mode is not `file`.

Closes #95

### DIFF
--- a/lua/telescope/_extensions/file_browser/picker.lua
+++ b/lua/telescope/_extensions/file_browser/picker.lua
@@ -37,7 +37,7 @@ local get_selection_index = function(opts, results)
       end
     end
   end
-  return 1
+  return vim.F.if_nil(opts.default_selection_index, 1)
 end
 
 --- List, create, delete, rename, or move files and folders of your cwd.
--- a/lua/telescope/_extensions/file_browser/picker.lua
+++ b/lua/telescope/_extensions/file_browser/picker.lua
@@ -30,16 +30,13 @@ local get_selection_index = function(opts, results)
   local buf_path = vim.api.nvim_buf_get_name(0)
   local current_dir = Path:new(buf_path):parent():absolute()
 
-  if opts.path ~= current_dir then
-    return 1
-  end
-
-  for i, path_entry in ipairs(results) do
-    if path_entry.value == buf_path then
-      return i
+  if opts.path == current_dir then
+    for i, path_entry in ipairs(results) do
+      if path_entry.value == buf_path then
+        return i
+      end
     end
   end
-
   return 1
 end
 
--- a/lua/telescope/_extensions/file_browser/picker.lua
+++ b/lua/telescope/_extensions/file_browser/picker.lua
@@ -67,6 +67,7 @@ end
 ---@field files boolean: start in file (true) or folder (false) browser (default: true)
 ---@field add_dirs boolean: whether the file browser shows folders (default: true)
 ---@field depth number: file tree depth to display, `false` for unlimited depth (default: 1)
+---@field select_buffer boolean: whether to select opened buffer if possible; may imply `hidden=true`. (default: false)
 ---@field hidden boolean: determines whether to show hidden files or not (default: false)
 ---@field respect_gitignore boolean: induces slow-down w/ plenary finder (default: false, true if `fd` available)
 ---@field browse_files function: custom override for the file browser (default: |fb_finders.browse_files|)
@@ -86,9 +87,9 @@ fb_picker.file_browser = function(opts)
   opts.hide_parent_dir = vim.F.if_nil(opts.hide_parent_dir, false)
   opts.select_buffer = vim.F.if_nil(opts.select_buffer, false)
 
-  local select_buffer = opts.select_buffer and opts.files and opts.depth == 1
+  local select_buffer = opts.select_buffer and opts.files
   -- handle case that current buffer is a hidden file
-  opts.hidden = (select_buffer and vim.fn.expand('%:p:t'):sub(1, 1) == '.') and true or opts.hidden
+  opts.hidden = (select_buffer and vim.fn.expand("%:p:t"):sub(1, 1) == '.') and true or opts.hidden
   local finder = fb_finder.finder(opts)
   -- find index of current buffer in the results
   if select_buffer then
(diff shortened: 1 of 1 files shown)


lua/telescope/_extensions/file_browser/picker.lua
```diff
@@ -48,6 +73,7 @@ local fb_picker = {}
 ---@field grouped boolean: group initial sorting by directories and then files; uses plenary.scandir (default: false)
 ---@field files boolean: start in file (true) or folder (false) browser (default: true)
 ---@field add_dirs boolean: whether the file browser shows folders (default: true)
+---@field select_buffer boolean: whether to select current buffer at start if possible (default: false)
```
fdschmidt93, line 76: Let's make clear this entails `hidden=true` if `select_buffer=true` and the current buffer is a hidden file

lua/telescope/_extensions/file_browser/picker.lua
```diff
@@ -25,6 +25,31 @@ local os_sep = Path.path.sep
...
+  local init_dir = Path:new(opts.path):absolute()
+  local buf_path = Path:new(vim.api.nvim_buf_get_name(0)):absolute()
+  local current_dir = Path:new(buf_path):parent():absolute()
+
+  if init_dir ~= current_dir then
+    return 1
+  end
+
+  for i, v in ipairs(results) do
+    local p = Path:new(v.value):absolute()
```
fdschmidt93, line 39: value should correspond to absolute paths; hence `Path:new` is not required here.

lua/telescope/_extensions/file_browser/picker.lua
```diff
@@ -25,6 +25,31 @@ local os_sep = Path.path.sep
...
+    if p == buf_path then
+      return i
+    end
+  end
+
+  return 1
+end
+
+-- test whether current buffer is a hidden file
+local buf_is_hidden_file = function()
```
fdschmidt93, line 49: No need to extract that into a separate function as it's only called once, just put it directly down below with a comment.

lua/telescope/_extensions/file_browser/picker.lua
```diff
@@ -25,6 +25,31 @@ local os_sep = Path.path.sep
 -- enclose in module for docgen
 local fb_picker = {}
 
+-- try to get the index of entry of current buffer
+local get_selection_index = function(opts, results)
+  local init_dir = Path:new(opts.path):absolute()
+  local buf_path = Path:new(vim.api.nvim_buf_get_name(0)):absolute()
```
fdschmidt93, line 31: `vim.api.nvim_buf_get_name` gets the full file name of the buffer which should correspond to `Path:new(...):absolute()`.

lua/telescope/_extensions/file_browser/picker.lua
```diff
@@ -25,6 +25,31 @@ local os_sep = Path.path.sep
 -- enclose in module for docgen
 local fb_picker = {}
 
+-- try to get the index of entry of current buffer
+local get_selection_index = function(opts, results)
+  local init_dir = Path:new(opts.path):absolute()
```
fdschmidt93, line 30: I'd expect `opt.path` to be an absolute path here at last, since it's already been `vim.fn.expand`ed.

lua/telescope/_extensions/file_browser/picker.lua
```diff
@@ -63,10 +89,23 @@ fb_picker.file_browser = function(opts)
...
   opts.path = opts.path and vim.fn.expand(opts.path) or opts.cwd
   opts.files = vim.F.if_nil(opts.files, true)
+  opts.select_buffer = vim.F.if_nil(opts.select_buffer, false)
+
+  -- process select_buffer option
+  local finder
+  if opts.files and opts.depth == 1 and opts.select_buffer then
+    if buf_is_hidden_file() then opts.hidden = true end
+    finder = fb_finder.finder(opts)
+    local results = finder:_browse_files().results
```
fdschmidt93, line 99: Right now, I'm pretty sure this means we are running `file_browser` effectively twice.

If you separate this into 

```lua
finder._finder = finder:_browse_files()
opts.default_selection_index = get_selection_index(opts, finder.results)
```

Does it still work?

lua/telescope/_extensions/file_browser/picker.lua
```diff
@@ -63,10 +89,23 @@ fb_picker.file_browser = function(opts)
   opts.cwd = opts.cwd and vim.fn.expand(opts.cwd) or cwd
   opts.path = opts.path and vim.fn.expand(opts.path) or opts.cwd
   opts.files = vim.F.if_nil(opts.files, true)
+  opts.select_buffer = vim.F.if_nil(opts.select_buffer, false)
+
+  -- process select_buffer option
+  local finder
+  if opts.files and opts.depth == 1 and opts.select_buffer then
```
fdschmidt93, line 96: Nitpick, let's change the order here from least to most likely

`opts.select_buffer and opts.files and opts.depth == 1`

The latter two might be argued, but I guess it's more often that users would want to start in `folder_browser` mode as they are unaware of `opts.depth`.

fdschmidt93 (commented):
Thanks for the PR :)

Overall looks good, only a few clean up nits.

fdschmidt93:
Apologies for the delay and thanks for the PR :)

PR #97, opened 2022-02-06
feat: add option `add_files`

severij:
I have this use case, where I want to be able to chain file_browser with other pickers like live_grep or find_files. The idea is that I could first select a folder using file_browser and then set the folder path to `cwd` in live_grep or find_files options. This option allows to only show folders when `add_files == false`. Of course, because `<CR>` opens a folder, some other key mapping is needed to actually select the folder.

### DIFF
--- a/doc/telescope-file-browser.txt
+++ b/doc/telescope-file-browser.txt
@@ -90,6 +90,8 @@ fb_picker.file_browser({opts})                      *fb_picker.file_browser()*
                                         browser (default: true)
         {add_dirs}          (boolean)   whether the file browser shows folders
                                         (default: true)
+        {add_files}         (boolean)   whether the file browser shows files
+                                        (default: true)
         {depth}             (number)    file tree depth to display, `false`
                                         for unlimited depth (default: 1)
         {dir_icon}          (string)    change the icon for a directory
--- a/lua/telescope/_extensions/file_browser/finders.lua
+++ b/lua/telescope/_extensions/file_browser/finders.lua
@@ -39,7 +39,11 @@ fb_finders.browse_files = function(opts)
     if opts.respect_gitignore == false then
       table.insert(args, "--no-ignore-vcs")
     end
-    if opts.add_dirs == false then
+    if opts.add_dirs then
+      table.insert(args, "--type")
+      table.insert(args, "directory")
+    end
+    if opts.add_files then
       table.insert(args, "--type")
       table.insert(args, "file")
     end
@@ -58,6 +62,7 @@ fb_finders.browse_files = function(opts)
   else
     local data = scan.scan_dir(opts.path, {
       add_dirs = opts.add_dirs,
+      add_files = opts.add_files,
       depth = opts.depth,
       hidden = opts.hidden,
     })
@@ -131,6 +136,7 @@ fb_finders.finder = function(opts)
     cwd = opts.cwd_to_path and opts.path or opts.cwd, -- nvim cwd
     path = vim.F.if_nil(opts.path, opts.cwd), -- current path for file browser
     add_dirs = vim.F.if_nil(opts.add_dirs, true),
+    add_files = vim.F.if_nil(opts.add_files, true),
     hidden = vim.F.if_nil(opts.hidden, false),
     depth = vim.F.if_nil(opts.depth, 1), -- depth for file browser
     respect_gitignore = vim.F.if_nil(opts.respect_gitignore, has_fd),
--- a/lua/telescope/_extensions/file_browser/picker.lua
+++ b/lua/telescope/_extensions/file_browser/picker.lua
@@ -48,6 +48,7 @@ local fb_picker = {}
 ---@field grouped boolean: group initial sorting by directories and then files; uses plenary.scandir (default: false)
 ---@field files boolean: start in file (true) or folder (false) browser (default: true)
 ---@field add_dirs boolean: whether the file browser shows folders (default: true)
+---@field add_files boolean: whether the file browser shows files (default: true)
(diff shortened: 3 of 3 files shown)


ikerurda:
This seems very similar to the `folder-browser` functionality, could you check if that satisfies your workflow?

Then you could use something like the following to open other pickers:
```lua
local builtin = require "telescope.builtin"
local open_in = function(finder)
  return function(prompt_bufnr)
    local path = action_state.get_selected_entry()[1]
    actions._close(prompt_bufnr, true)
    finder { cwd = path }
  end
end

-- ...
-- then map it to some key:
mappings = {
  i = {
    ["<a-f>"] = open_in(builtin.find_files),
    ["<a-g>"] = open_in(builtin.live_grep),
  },
},
```

severij:
> This seems very similar to the `folder-browser` functionality, could you check if that satisfies your workflow?
Thanks, I think I can achieve most of the things I want with the solution you described. Still, often I may or may not know the name of the folder I'm looking for, so file_browser is more useful in that case. I think it would be nice to be able to toggle `hide/show files` in file_browser, whenever you have a lot of files but few folders in the results.

fdschmidt93:
I'm okay enabling this use case. Since this is a very individual use case, I wouldn't hit the merge button just yet as `telescope-file-browser` doesn't rapidly change anymore anyways and let you work off of your fork if that's ok for the time being.

What I'd prefer doing:

1. We probably need to have some separation between `{file, folder}_browser` arguments because they'll probably just keep on growing
2. I feel like this should be the equivalent of `folder_browser` with `depth=1` and `cwd_to_path=true` (maybe we can rename this..) to keep `file_` and `folder_browser` adequately logically separated
3. We should probably try to refactor `plenary.scandir` and `fd` flags into some sort of factory function with harmonized flags via metatables or something, assuming that can be done cleanly. The idea would be any user can then opt-in to any flag, maybe generate any toggling action easily, etc.

If we have 1 and 3 this would probably follow very naturally. I would probably try and work on that this week.

severij:
Setting `depth=1` and `cwd_to_path=true` for folder_browser doesn't quite still fill my needs. It would be nice to be able to stay in `Folder Browser` so that whenever a folder is selected in that mode it doesn't throw me back to `File Browser`. Also I would like to be able to go back to the parent folder with `../` while staying in `Folder browser`. In my fork `add_files=false` allows me to have this exact functionality, although in `File browser`.

To me folder_browser seems like a special case of file_browser, where `hide_parent_dir=true`, `add_dirs=true`, `add_files=false` (introduced in this PR) and `depth=false`, so I don't quite understand why file_bowser and folder_browser are kept logically separate. To me the current operating model of folder_browser (fuzzy finding with no max_depth) seems to be quite distinct from file_browser where you can browse the folder hierarchy like with any other file explorer.

But yeah, I'm okay working with my current fork for the time being. I'm thinking of remapping `<C-f>` to cycle between the options *show files and folders* (`add_dirs=true` and `add_files=true`), *show folders only*  (`add_dirs=true` and `add_files=false`) and *show files only* (`add_dirs=false` and `add_files=true`), and add those mappings that @ikerurda suggested.

fdschmidt93:
> It would be nice to be able to stay in Folder Browser so that whenever a folder is selected in that mode it doesn't throw me back to File Browser.

Maybe I haven't thought your use case perfectly through, but so long as there is no toggling added to your functionality (which ofc this PR lacks), you wouldn't have really done all too much apart from grepping / launching another picker with the selected folder to begin with (with `add_files=false`) as any folder you follow in `file_browser` will always only show folders as well? Though I guess that's something you would have ended up adding. So I thought of it more like you'd want a special one-time view of folders, as in practice you'd launch file or folder browser to select folders to grep then I guess.

In any case, workflows are a very individual thing and I'm happy to try to facilitate yours :) 

I'll try to implement 1 and 3 this week in a more harmonious way that the user can seamlessly add any `fd` or `plenary` flags and possible create his own actions on the fly. If I understand it correctly, it should then ideally boil down to passing the right flag to the picker and constructing the action to (similar to `telescope.actions.generate`). Maybe as to why I go to such lengths -- it's primarily about maintainability: adding/changing/whatnot about flags to `plenary.scandir` and `fd` came / will come up frequently enough that I'd want a more permanent solution.

severij:
> Maybe I haven't thought your use case perfectly through, but so long as there is no toggling added to your functionality (which ofc this PR lacks), you wouldn't have really done all too much apart from grepping / launching another picker with the selected folder to begin with (with add_files=false) as any folder you follow in file_browser will always only show folders as well? Though I guess that's something you would have ended up adding. So I thought of it more like you'd want a special one-time view of folders, as in practice you'd launch file or folder browser to select folders to grep then I guess.

Effectively what I wanted to add with this PR was the ability to hide files in the file_browser. In my work I often have hundres of files showing in the `Results` window and only few folders, so it can be quite challenging if I want to browse just the folders. The ability to create a mapping for toggling `show/hide files` would help in those scenarios. In my original message I tried to come up some examples where this would be useful but I understand now that it might've just confused you more with the examples. Sorry for not being clear enough. 😅

> I'll try to implement 1 and 3 this week in a more harmonious way that the user can seamlessly add any fd or plenary flags and possible create his own actions on the fly. If I understand it correctly, it should then ideally boil down to passing the right flag to the picker and constructing the action to (similar to telescope.actions.generate). Maybe as to why I go to such lengths -- it's primarily about maintainability: adding/changing/whatnot about flags to plenary.scandir and fd came / will come up frequently enough that I'd want a more permanent solution.

Yes, allowing one to add arbitrary flags to `plenary.scandir` and `fd` makes more sense, now that you said it. It allows me to achieve the functionality I wanted and much more. Thank you for your effort, I'll close this PR. 🙂

PR #107, opened 2022-03-08
Fixes problem with using the wrong check to determine if on Windows or not

waynebowie99:
win-32 always return 0 on windows as the correct check is win32. This fixes the problem I ran into with not being able to use the open function


fdschmidt93:
Thanks :)

PR #111, opened 2022-03-16
feat: enable hijacking of netrw on vim startup

jamestrew:
There's been several requests of launching `file_browser` on `VimEnter` over netwr. I was initially skeptical of implementing this but considering this plugin is somewhat an alternative to file browsers/explorers/trees like [nvim-tree](https://github.com/kyazdani42/nvim-tree.lua) and seeing as how they've implemented it, I thought why not. At least as an opt-in feature.

I can add some notes in the docs/readme but I wanted to get some thoughts on the idea as well as the potentially over-simplistic implementation of it.

### DIFF
--- a/doc/telescope-file-browser.txt
+++ b/doc/telescope-file-browser.txt
@@ -96,48 +96,54 @@ fb_picker.file_browser({opts})  *telescope-file-browser.picker.file_browser()*
         {opts} (table)  options to pass to the picker
 
     Fields: ~
-        {path}              (string)      dir to browse files from from,
-                                          `vim.fn.expanded` automatically
-                                          (default: vim.loop.cwd())
-        {cwd}               (string)      dir to browse folders from,
-                                          `vim.fn.expanded` automatically
-                                          (default: vim.loop.cwd())
-        {cwd_to_path}       (boolean)     whether folder browser is launched
-                                          from `path` rather than `cwd`
-                                          (default: false)
-        {grouped}           (boolean)     group initial sorting by directories
-                                          and then files; uses plenary.scandir
-                                          (default: false)
-        {files}             (boolean)     start in file (true) or folder
-                                          (false) browser (default: true)
-        {add_dirs}          (boolean)     whether the file browser shows
-                                          folders (default: true)
-        {depth}             (number)      file tree depth to display, `false`
-                                          for unlimited depth (default: 1)
-        {select_buffer}     (boolean)     select current buffer if possible;
-                                          may imply `hidden=true` (default:
-                                          false)
-        {hidden}            (boolean)     determines whether to show hidden
-                                          files or not (default: false)
-        {respect_gitignore} (boolean)     induces slow-down w/ plenary finder
-                                          (default: false, true if `fd`
-                                          available)
-        {browse_files}      (function)    custom override for the file browser
-                                          (default: |fb_finders.browse_files|)
-        {browse_folders}    (function)    custom override for the folder
-                                          browser (default:
-                                          |fb_finders.browse_folders|)
-        {hide_parent_dir}   (boolean)     hide `../` in the file browser
-                                          (default: false)
-        {quiet}             (boolean)     surpress any notification from
-                                          file_brower actions (default: false)
-        {dir_icon}          (string)      change the icon for a directory
-                                          (default: )
-        {dir_icon_hl}       (string)      change the highlight group of dir
-                                          icon (default: "Default")
-        {display_stat}      (bool|table)  ordered stat; see above notes,
-                                          (default: `{ date = true, size =
-                                          true }`)
+        {path}              (string)         dir to browse files from from,
+                                             `vim.fn.expanded` automatically
+                                             (default: vim.loop.cwd())
+        {cwd}               (string)         dir to browse folders from,
+                                             `vim.fn.expanded` automatically
+                                             (default: vim.loop.cwd())
+        {cwd_to_path}       (boolean)        whether folder browser is
+                                             launched from `path` rather than
+                                             `cwd` (default: false)
+        {grouped}           (boolean)        group initial sorting by
+                                             directories and then files; uses
+                                             plenary.scandir (default: false)
+        {files}             (boolean)        start in file (true) or folder
+                                             (false) browser (default: true)
+        {add_dirs}          (boolean)        whether the file browser shows
+                                             folders (default: true)
+        {depth}             (number)         file tree depth to display,
+                                             `false` for unlimited depth
+                                             (default: 1)
+        {select_buffer}     (boolean)        select current buffer if
+                                             possible; may imply `hidden=true`
+                                             (default: false)
+        {hidden}            (boolean)        determines whether to show hidden
+                                             files or not (default: false)
+        {respect_gitignore} (boolean)        induces slow-down w/ plenary
+                                             finder (default: false, true if
+                                             `fd` available)
+        {browse_files}      (function)       custom override for the file
+                                             browser (default:
+                                             |fb_finders.browse_files|)
+        {browse_folders}    (function)       custom override for the folder
+                                             browser (default:
+                                             |fb_finders.browse_folders|)
+        {hide_parent_dir}   (boolean)        hide `../` in the file browser
+                                             (default: false)
+        {quiet}             (boolean)        surpress any notification from
+                                             file_brower actions (default:
+                                             false)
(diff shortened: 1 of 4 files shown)


fdschmidt93:
I've checked out this PR and get a stackoverflow when `hijack_netrw` is `true`. Could be due to some deferred loading of telescope, which I've tried to deactivate but causes some further issues with packer. Does it work for you out-of-the-box?

lua/telescope/_extensions/file_browser.lua
```diff
@@ -148,6 +143,18 @@ local file_browser = function(opts)
...
 end
 
+local fb_setup = function(opts)
+  -- TODO maybe merge other keys as well from telescope.config
+  pconf.mappings = vim.tbl_deep_extend("force", pconf.mappings, require("telescope.config").values.mappings)
+  pconf = vim.tbl_deep_extend("force", pconf, opts)
+
+  if pconf.hijack_netrw then
+    vim.cmd "silent! autocmd! FileExplorer *"
+    vim.cmd "autocmd VimEnter * ++once silent! autocmd! FileExplorer *"
```
fdschmidt93, line 153: So how is netrw supposed to work here? IIRC it only opens when you open a folder, right? e.g. `nvim .` or `nvim some_folder` rather than `nvim`.

I don't think file browser should always be opened on `VimEnter`.

jamestrew:
> Does it work for you out-of-the-box?

It was initially, but not anymore. I must've changed something... I will make some tweaks.

lua/telescope/_extensions/file_browser.lua
jamestrew: You're right, I'll add tweak this as well

jamestrew:
Should be functional now. But lazy loading telescope does seem to cause some issues so it's not recommended. Let me know if you experience any issues besides that.

I'd like to add some docs on this but it didn't feel quite right to add this option in the docstring for `fb_picker.file_browser` as it's more of one time initialization setting option rather than a picker option.

lua/telescope/_extensions/file_browser.lua
```diff
@@ -173,6 +169,22 @@ local file_browser = function(opts)
   fb_picker(popts)
 end
 
+local fb_setup = function(opts)
+  -- TODO maybe merge other keys as well from telescope.config
+  pconf.mappings = vim.tbl_deep_extend("force", pconf.mappings, require("telescope.config").values.mappings)
+  pconf = vim.tbl_deep_extend("force", pconf, opts)
+
+  if pconf.hijack_netrw then
+    vim.cmd "silent! autocmd! FileExplorer *"
```
fdschmidt93, line 178: - switch to new autocmd api
- `nvim lua/` in file browser should open `file browser` in that folder analogous to netrw
- Didn't we discuss to not open `file_browser` by default on `VimEnter`? It should properly follow netrw and only trigger file browser in `nvim .` where `.` indicates `$CWD`

fdschmidt93:
> I'd like to add some docs on this but it didn't feel quite right to add this option in the docstring for fb_picker.file_browser as it's more of one time initialization setting option rather than a picker option.

I guess `fb_setup` then should get a separate section akin to how it works for `fb_actions`, `fb_picker.

ranebrown:
Can this be made to also work on `:e some/dir/`?

jamestrew:
> Can this be made to also work on `:e some/dir/`?

Yes I'm planning to work that into this PR as well eventually. Currently holding off on doing too much work until #118 is merged.

ranebrown:
Dropping this here in case it is useful. Seems to be working fairly well so far.

```lua
-- https://github.com/nvim-telescope/telescope-file-browser.nvim/blob/f68d39baf90856678301a3310b35a71d59f5d456/lua/telescope/_extensions/file_browser/make_entry.lua#L49
local get_fb_prompt = function()
    local prompt_bufnr = vim.tbl_filter(function(b)
        return vim.bo[b].filetype == "TelescopePrompt"
    end, vim.api.nvim_list_bufs())
    -- vim.ui.{input, select} might be telescope pickers
    if #prompt_bufnr > 1 then
        for _, buf in ipairs(prompt_bufnr) do
            local current_picker =
                require("telescope.actions.state").get_current_picker(
                    prompt_bufnr
                )
            if current_picker.finder._browse_files then
                prompt_bufnr = buf
                break
            end
        end
    else
        prompt_bufnr = prompt_bufnr[1]
    end
    return prompt_bufnr
end

vim.g.loaded_netrw = 1
vim.g.loaded_netrwPlugin = 1
vim.api.nvim_create_augroup("FileExplorer", { clear = true })
vim.api.nvim_create_autocmd({ "BufEnter", "BufNewFile" }, {
    group = "FileExplorer",
    pattern = { "*" },
    callback = function()
        vim.schedule(function()
            local tp = get_fb_prompt()
            if vim.fn.bufexists(tp) == 1 then
                return
            end

            local buf = vim.api.nvim_get_current_buf()
            local bufname = vim.api.nvim_buf_get_name(buf)

            -- prevents reopening file-browser if exiting without selecting a file
            if vim.g.fb_bufname == bufname then
                -- clear the global var after short time so the same folder can be reopened
                vim.defer_fn(function()
                    vim.g.fb_bufname = ""
                end, 1000)
                return
            else
                vim.g.fb_bufname = bufname
            end

            if vim.fn.isdirectory(bufname) ~= 1 then
                return
            else
                require("telescope").extensions.file_browser.file_browser({
                    cwd = vim.fn.expand("%:p:h"),
                })
            end
        end)
    end,
    desc = "Open Telescope file-browser for directories",
})

```

jamestrew:
> Dropping this here in case it is useful. Seems to be working fairly well so far.

Thanks it looks solid and I'll probably end up using it. 

I'm curious why this is necessary. Specifically the deferring for 1000ms. 
```lua
 vim.defer_fn(function()
      vim.g.fb_bufname = ""
  end, 1000)
```
Having been testing your snippet for a while, I can't seem to find a use case where deferring makes any difference over just straight up calling `vim.g.fb_bufname = ''`.

ranebrown:
Hmm, thought it was necessary for repoening the same folder e.g. `:e folder/` exit file-browser followed by `:e folder/` but you are right - it seems to work without it. 🤷

doc/telescope-file-browser.txt
```diff
@@ -96,48 +96,54 @@ fb_picker.file_browser({opts})  *telescope-file-browser.picker.file_browser()*
...
+                                             file_brower actions (default:
+                                             false)
+        {dir_icon}          (string)         change the icon for a directory
+                                             (default: )
+        {dir_icon_hl}       (string)         change the highlight group of dir
+                                             icon (default: "Default")
+        {display_stat}      (boolean|table)  ordered stat; see above notes,
+                                             (default: `{ date = true, size =
+                                             true }`)
+        {hijack_netrw}      (boolean)        use telescope file browser when
```
fdschmidt93, line 144: Docs are autogenerated from lua files, so the changed would have to be made there. For the time being, let's maybe document this on the README somewhere, as the corresponding docs for `config.lua` are not yet generated and are also not so easy to generate appropriately. `telescope.nvim` does that in quite an involved way.

lua/telescope/_extensions/file_browser/config.lua
```diff
@@ -62,6 +62,39 @@ _TelescopeFileBrowserConfig = {
...
+    callback = function()
+      vim.schedule(function()
+        local bufname = vim.api.nvim_buf_get_name(0)
+        if vim.fn.isdirectory(bufname) == 0 then
+          config.values.netrw_bufname = vim.fn.expand "#:p:h"
+          return
+        end
+
+        -- prevents reopening of file-browser if exiting without selecting a file
+        if config.values.netrw_bufname == bufname then
```
fdschmidt93, line 82: Can't this be just a local variable outside the `hijack` function? It's not part of the config as such and ideally shouldn't be.

fdschmidt93 (approved):
LGTM, thanks :)

The last thing that comes to my mind is as to whether you would like to mention the feature on the README somewhere -- not strictly required.. users should look into vimdocs, we make great effort that hopefully all relevant stuff is sufficiently documented.

I'd merge either on your sign-off, i.e. either you're fine not mentioning it or add it somewhere on README.

jamestrew:
:+1: 
added a line in the readme + removed some trailing whitespaces

fdschmidt93 (approved):
Thanks :)

PR #115, opened 2022-03-22
fix: preserve custom titles

Zorbn:
Telescope allows the user to specify "prompt_title" and "results_title". I was making a theme and used those values to have custom titles which works fine for the default finders, but not for this one because the titles change upon changing directory. This fix prevents "fb_utils.redraw_border_title" from overwriting the titles if they are custom.

### DIFF
--- a/lua/telescope/_extensions/file_browser/finders.lua
+++ b/lua/telescope/_extensions/file_browser/finders.lua
@@ -149,6 +149,8 @@ fb_finders.finder = function(opts)
     close = function(self)
       self._finder = nil
     end,
+    custom_prompt_title = opts.custom_prompt_title,
+    custom_results_title = opts.custom_results_title,
   }, {
     __call = function(self, ...)
       -- (re-)initialize finder on first start or refresh due to action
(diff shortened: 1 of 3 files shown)


lua/telescope/_extensions/file_browser/finders.lua
```diff
@@ -149,6 +149,8 @@ fb_finders.finder = function(opts)
     close = function(self)
       self._finder = nil
     end,
+    custom_prompt_title = opts.custom_prompt_title,
```
fdschmidt93, line 152: For simplicity, I think we can just call this `{prompt, results}_title`? They'll only ever be set for custom titles.

fdschmidt93 (commented):
Slight nit, but otherwise LGTM :)

lua/telescope/_extensions/file_browser/finders.lua
Zorbn: Agreed, I simplified the names as you suggested.

fdschmidt93:
Thanks :)

PR #117, opened 2022-03-25
feat: create & open file/folder from prompt if results empty

### LINKED ISSUE #114: Create new file when file dont exists
Would be nice to have an option when i type the name of file:

i type:
>> my_file.js

On telescope:
__________
[+] my_file.js (new file)
../
__________

and when i select this option, it create a file and opens on editor.

Emacs do it and its awesome.

fdschmidt93:
Closes #114

cc @danielnehrig this also might be something you're looking for

This PR enables creating a file or folder in file browser mode on select default (i.e. `Enter`) whenever a prompt does not yield a result (i.e. results are empty). For the time being, this is the new default behavior, as 99% of cases users shouldn't be hitting enter randomly on empty results.

If anyone ever complains, feel free to provide a PR with possibility to opt out :) I just wanted to have fun with a quick & cool feature. Enabling opting out is the not so fun part :sweat_smile: 

Showcase

https://user-images.githubusercontent.com/39233597/160089672-3d62abc7-6949-43f9-ab02-39ff1ae217ef.mp4

### DIFF
--- a/lua/telescope/_extensions/file_browser.lua
+++ b/lua/telescope/_extensions/file_browser.lua
@@ -56,6 +56,8 @@ local fb_utils = require "telescope._extensions.file_browser.utils"
 
 local action_state = require "telescope.actions.state"
 local action_set = require "telescope.actions.set"
+local state = require "telescope.state"
+local Path = require "plenary.path"
 
 local pconf = {
   mappings = {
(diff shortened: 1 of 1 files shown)


fdschmidt93:
I'm feeling frisky. I'll just merge this as 3-5mins of trial of the feature indicate it works as expected and worst case scenario an extra file/folder should be created.

danielnehrig:
thanks!

Issue #87, opened 2022-01-26
Having to confirm twice on operation cancel 


tami5:
### Description

This is annoying a bit. the second line printed makes the prompt stay there instead of just be printed in console.
![Screen Shot 2022-01-26 at 3 04 36 PM](https://user-images.githubusercontent.com/65782666/151159978-ed5e5810-c6d4-4339-a782-936510686d04.png)

maybe differing results would make sense and solve this?

### Neovim version

```markdown
NVIM v0.6.1
```

### Operating system and version

macOs

### Steps to reproduce

1. create/rename/delete
2. cancel or confirm 
3. console await user input

### Expected behavior

after confirming or canceling action the focus should be back on the browser

### Actual behavior

Await additional input.

### Minimal config

```Lua
..
```

fdschmidt93:
I'd hope that this would be resolved by switching towards notifications?

#72 

I have to setup notifications and take a look at that PR.

tami5:
I guess it would be a by product of #72.

jamestrew:
Unfortunately #72 doesn't solve this `Press ENTER or type command to continue` issue.

I'll look into it a bit more within the next couple of days. All I know is that stacking too many multline `vim.notify` or `print` beyond the `cmdheight` is the culprit. I'm not super sure how to force clear what's in the command bar besides throwing a `vim.api.nvim_command('norm :esc<CR>')` between the `vim.notify`s which feels very janky and dumb.

tami5:
Hmmm maybe then no new lines? flat text

fdschmidt93:
So I've been cleaning up my dots a bit and also finally got around to installing `nvim-notify` and playing with that a bit.

I'll probably take over #72 eventually with my own PR (though borrowing some ideas :), from what I can tell there's two-three options here:
- swapping `vim.notify` with `nvim-notify`
- increasing `cmdheight` intermittently (a bit messy)
- Put messages into 1 line (as will my refactor) and then `vim.cmd [[ redraw! ]]` seems to work such that the 1 line solely remains in `cmdline`

As I understand it, `hit-enter` prompt is essentially triggering the redraw as it's otherwise not clear for `(n)vim` how to handle intermediate excess `cmdheight`.

fdschmidt93:
This should hopefully be addressed now, so closing :)

Please let me know otherwise.

PR #118, opened 2022-03-26
feat: notify and vim.ui harmonization

### LINKED ISSUE #91: Add "quiet" config option to suppress action confirmation print statements

Whenever I execute a file action (eg: delete a file), I find the additional confirmation at the end of the action, which then requires an additional enter to dismiss, breaks my workflow

eg:
```
These files are going to be deleted:
/tmp/foo
Remove selected files [y/N]: y
/tmp/foo has been removed!                             \ these 2 lines
Press ENTER or type command to continue                / 
```

ie: the following print statements in `actions.lua`

```
print(string.format("%s has been moved!", filename))
print(string.format("\n%s has been removed!", p:absolute()))
```

I would like a "quiet" config option which prevents the action confirmation print statement being output

fdschmidt93:
Continuation of #72

Closes #91 #87 

Changelog:
- Use `vim.ui.input` and `vim.notify` everywhere
- Smoothly pretty much all `hit-enter prompt`s by `redraw` on `vim.ui.input` with default `vim.notify` and casting all messages on a single line (probably not amazing if tons of single files/folders are created/deleted/etc. but can deal with it later)
- Introduce `quiet` option to suppress all outputs 
- Refactoring a proper config module
- Disallow removing parent folder (`file_browser`) and current folder (`folder_browser`) (yes, I've done something like that accidentally in my testing.. all my lua repos had to be restored via timeshift haha)
 
Since it is a very large PR, I apologize in before for any regression caused :) In my testing I couldn't find any, I've been on this branch some time and retested important actions repeatedly, but I'm sure a larger user base will find things.

### DIFF
--- a/.github/workflows/docgen.yml
+++ b/.github/workflows/docgen.yml
@@ -36,9 +36,7 @@ jobs:
         run: |
           # We have to build the parser every single time to keep up with parser changes
           cd ~/.local/share/nvim/site/pack/vendor/start/tree-sitter-lua
-          mkdir -p build parser
-          cc -o ./build/parser.so -I./src src/parser.c src/scanner.cc -shared -Os -lstdc++ -fPIC
-          ln -s ../build/parser.so parser/lua.so
+          make dist
           cd -
 
       - name: Generating docs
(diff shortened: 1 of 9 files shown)


fdschmidt93:
@ikerurda may I kindly ask you to test drive this PR please? :)

PR #130, opened 2022-05-12
feat: create file from prompt + create fixes

fdschmidt93:
This PR introduces the `fb_actions.create_from_prompt` which allows to create files or folder from prompt and directly open them via `telescope-file-browser.nvim` (i.e. if you open a folder, you'll land in that folder).

Creds to @otavioschwanck for the original idea and implementation :)

Furthermore, this commit reflects a breaking change. The input callback to create a file or folder via the prompt if no result is shown is now superseded by the `fb_actions.create_from_prompt` mapped by default to Shift-CR (works if your terminal recognizes this, otherwise please remap).

cc @danielnehrig @ZhiyuanLck as you might have used that feature

Please don't be too concerned with the default mapping; once some previewer upstream refactoring etc is committed, I plan to reconsider mappings and disseminate info about recent file browser updates on reddit to draw attention to a wider audience.

### DIFF
--- a/doc/telescope-file-browser.txt
+++ b/doc/telescope-file-browser.txt
@@ -169,11 +169,25 @@ require('telescope').setup {
 <
 
 fb_actions.create({prompt_bufnr})    *telescope-file-browser.actions.create()*
-    Creates a new file in the current directory of the
+    Creates a new file or dir in the current directory of the
     |telescope-file-browser.picker.file_browser|.
     - Finder:
       - file_browser: create a file in the currently opened directory
       - folder_browser: create a file in the currently selected directory
+    - Notes:
+      - You can create folders by ending the name in the path separator of your
+        OS, e.g. "/" on Unix systems
+      - You can implicitly create new folders by passing
+        $/CWD/new_folder/filename.lua
+
+
+    Parameters: ~
+        {prompt_bufnr} (number)  The prompt bufnr
+
+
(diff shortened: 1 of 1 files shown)


danielnehrig:
might it not make sens to make the default behaviour something else then `<S-CR> pressing shift+cr does nothing for me probably as you mentioned i have to set some thing in my kitty conf

otavioschwanck:
> might it not make sens to make the default behaviour something else then ` pressing shift+cr does nothing for me probably as you mentioned i have to set some thing in my kitty conf

i remapped to Ctrl + space on my config.

otavioschwanck:
> > might it not make sens to make the default behaviour something else then ` pressing shift+cr does nothing for me probably as you mentioned i have to set some thing in my kitty conf
> 
> i remapped to Ctrl + space on my config.  Would be nice to have something like that by default

fdschmidt93:
> Please don't be too concerned with the default mapping; once some previewer upstream refactoring etc is committed, I plan to reconsider mappings and disseminate info about recent file browser updates on reddit to draw attention to a wider audience.

Please understand I won't introduce a single mapping breaking change to a perfectly logical mapping that for some reason does not work for you. As per above, mappings as a whole will be reviewed eventually. Remapping is for good reason very easy -- we won't be able to please everyone, since there are so many mappings in the file browser.

PR #132, opened 2022-05-14
fix: pass defaults thru theme

jamestrew:
Currently, using the non-default picker theme won't pass your defaults through correctly forcing the plugin defaults to be used.


fdschmidt93 (approved):
LGTM, thanks :)

PR #134, opened 2022-05-16
rf(entry_maker): improve performance

fdschmidt93:
Currently, `telescope-file-browser` was always creating a `plenary.path` object for each entry all the time (artifact from old code).

Since `fd`, `plenary.scandir` or provider of choosing has to feed in absolute paths, we should be able to get rid of that.

Additionally, all table attributes apart from `value` and `ordinal` should be lazy. The improvements vis-a-vis `master` are best seen when launching folder browser from `/home/$USER/`.


fdschmidt93:
Ok, let's just merge this. I'm reasonably optimistic the diff won't introduce a regression and should be strictly beneficial :)

Once https://github.com/nvim-telescope/telescope.nvim/pull/1491 lands, the folder browser will be very usable from `/home/$USER/` (in my case 100K folders).

PR #139, opened 2022-05-18
fix(entry_maker): stat may be unavailable

fdschmidt93:
Closes #129 

Does that fix the issues for you @raj-magesh and @miversen33 ?

Seems to work well for the original issue for me. Primary question now to some degree for me is whether the broken symlink should be shown to begin with as it's not an appropriate file.

`fd` at best with `--show-errors` highlights the broken symlink in red.

PR #140, opened 2022-05-19
Added different formatting option for missing stat

miversen33:
Adds handling around if entry.stat doesn't exist, additionally logs out that the file is missing.

### DIFF
--- a/lua/telescope/_extensions/file_browser/make_entry.lua
+++ b/lua/telescope/_extensions/file_browser/make_entry.lua
@@ -14,9 +14,6 @@ local SIZE = {
   width = 7,
   right_justify = true,
   display = function(entry)
-    if not entry.stat then
-      log.debug("" .. entry.Path:absolute() .. " appears to be an invalid symlink")
-      return '0.0' end
     local size = entry.stat.size
     -- TODO(conni2461): If type directory we could just return 4.0K
     for _, v in ipairs(SIZE_TYPES) do
@@ -39,9 +36,6 @@ local DATE = {
   width = 13,
   right_justify = true,
   display = function(entry)
-    if not entry.stat then
-      log.debug("" .. entry.Path:absolute() .. " appears to be an invalid symlink")
-      return 'NA' end
     local mtime = entry.stat.mtime.sec
     if YEAR ~= os.date("%Y", mtime) then
       return os.date("%b %d  %Y", mtime)
--- a/lua/telescope/_extensions/file_browser/make_entry.lua
+++ b/lua/telescope/_extensions/file_browser/make_entry.lua
@@ -190,7 +190,9 @@ local make_entry = function(opts)
     if k == "stat" then
       local stat = vim.loop.fs_stat(t.value)
       t.stat = vim.F.if_nil(stat, false)
-      if not t.stat then log.debug("Unable to get stat for " .. t.value) end
+      if not t.stat then
+        log.debug("Unable to get stat for " .. t.value)
+      end
       return stat
     end
 
--- a/lua/telescope/_extensions/file_browser/make_entry.lua
+++ b/lua/telescope/_extensions/file_browser/make_entry.lua
@@ -37,6 +37,8 @@ local DATE = {
   right_justify = true,
   display = function(entry)
     local mtime = entry.stat.mtime.sec
+    -- Not really a great way to account for invalid mtimes but this works
+    if mtime < 0 then return 'NA' end
     if YEAR ~= os.date("%Y", mtime) then
       return os.date("%b %d  %Y", mtime)
     end
--- a/lua/telescope/_extensions/file_browser/make_entry.lua
+++ b/lua/telescope/_extensions/file_browser/make_entry.lua
@@ -188,6 +188,21 @@ local make_entry = function(opts)
     end
     if k == "stat" then
       local stat = vim.loop.fs_stat(t.value)
+      if not stat then
+        -- Kinda ick but since stat is a table of tables (and stuff),
+        -- we have to account for those subtables
+        log.warn("Unable to get stat for " .. t.Path:absolute())
+        stat = {}
+        setmetatable(stat, {
+          __index = function(_t, _k)
+            if
+              _k == 'atime'
+              or _k == 'birthtime'
+              or _k == 'mtime' then
+                return {nsec=-1, sec=-1}
+              end
+          return 0 end})
+      end
       t.stat = stat
       return stat
     end
--- a/lua/telescope/_extensions/file_browser/make_entry.lua
+++ b/lua/telescope/_extensions/file_browser/make_entry.lua
@@ -37,8 +37,6 @@ local DATE = {
   right_justify = true,
   display = function(entry)
     local mtime = entry.stat.mtime.sec
-    -- Not really a great way to account for invalid mtimes but this works
-    if mtime < 0 then return 'NA' end
     if YEAR ~= os.date("%Y", mtime) then
       return os.date("%b %d  %Y", mtime)
     end
@@ -152,12 +150,15 @@ local make_entry = function(opts)
     if #path_display > opts.file_width then
       path_display = truncate(path_display, opts.file_width, nil, -1)
     end
-    table.insert(display_array, path_display)
+    table.insert(display_array, entry.stat and path_display or { path_display, "WarningMsg" })
     table.insert(widths, { width = opts.file_width })
     if opts.display_stat then
       for _, v in pairs(opts.display_stat) do
-        table.insert(widths, { width = v.width, right_justify = v.right_justify })
-        table.insert(display_array, { v.display(entry), v.hl })
+        -- stat may be false meaning file not found / unavailable, e.g. broken symlink
+        if entry.stat then
+          table.insert(widths, { width = v.width, right_justify = v.right_justify })
+          table.insert(display_array, { v.display(entry), v.hl })
+        end
       end
     end
     local displayer = entry_display.create {
@@ -188,22 +189,8 @@ local make_entry = function(opts)
     end
     if k == "stat" then
       local stat = vim.loop.fs_stat(t.value)
-      if not stat then
-        -- Kinda ick but since stat is a table of tables (and stuff),
-        -- we have to account for those subtables
-        log.warn("Unable to get stat for " .. t.Path:absolute())
-        stat = {}
-        setmetatable(stat, {
-          __index = function(_t, _k)
-            if
-              _k == 'atime'
-              or _k == 'birthtime'
-              or _k == 'mtime' then
-                return {nsec=-1, sec=-1}
-              end
-          return 0 end})
-      end
-      t.stat = stat
+      t.stat = vim.F.if_nil(stat, false)
+      if not t.stat then log.debug("Unable to get stat for " .. t.value) end
       return stat
     end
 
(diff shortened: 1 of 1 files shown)


lua/telescope/_extensions/file_browser/make_entry.lua
```diff
@@ -13,6 +14,9 @@ local SIZE = {
   width = 7,
   right_justify = true,
   display = function(entry)
+    if not entry.stat then
```
fdschmidt93, line 17: I think this be would logging twice for a broken symlink as is (assuming you're both showing date and size). Let's move logging to the `stat` metatable key if it's not found :) Then we can also universally catch existence of `stat` in the loop in which it is added to displayed items similar to what I did (so no particular function has to check `stat` is available, esp. since I eventually want to allow users to hook into that.)

I'm also not sure whether invalid symlink would be the only cause of having no `stat` to begin with. Unless we're certain, let's have a more universal logging message.

lua/telescope/_extensions/file_browser/make_entry.lua
```diff
@@ -13,6 +14,9 @@ local SIZE = {
...
+      log.debug("" .. entry.Path:absolute() .. " appears to be an invalid symlink")
+      return '0.0' end
```
fdschmidt93, line 19: I think this should be just a `0` since there is no division.

fdschmidt93 (commented):
Yeah, logging is great, we should do that :)

I'll reply here --

> fd at best with --show-errors highlights the broken symlink in red.

Sorry, too little context. It shows the broken symlink red if you run it in your shell (not yet `telescope-file-browser.nvim`).

We can incorporate this quite nicely with

```lua
table.insert(display_array, entry.stat and path_display or { path_display, "WarningMsg" })
```
on line 152 which I think is nice visual feedback in conjunction with `"NA"` and `0`.

lua/telescope/_extensions/file_browser/make_entry.lua
miversen33: > I think this be would logging twice for a broken symlink as is (assuming you're both showing date and size).
You're right it definitely was. I have migrated the logging into a metatable.
**Note:** My experience with Lua is quite lacking so metatables are a bit of blackmagic to me. Lemme know if I can do this better.

lua/telescope/_extensions/file_browser/make_entry.lua
miversen33: Agreed, corrected

lua/telescope/_extensions/file_browser/make_entry.lua
```diff
@@ -185,6 +188,21 @@ local make_entry = function(opts)
     end
     if k == "stat" then
       local stat = vim.loop.fs_stat(t.value)
+      if not stat then
```
fdschmidt93, line 191: I feel that's unnecessarily unwieldy.

What do you think about my original solution with simply bypassing file size and date for files w/o stat: https://github.com/nvim-telescope/telescope-file-browser.nvim/pull/139/files

I'd say we add logging to that and highlight files with no stat (for which something is wrong) with

```lua
table.insert(display_array, entry.stat and path_display or { path_display, "WarningMsg" })
```

as when there's not stat, there's not really anything meaningful (ie size, date, ...) to show anyways, or is there?

lua/telescope/_extensions/file_browser/make_entry.lua
fdschmidt93: Metatables (MT) in brief on the example of the entry maker.

MT define a certain set of meta methods (`__index, __newindex, __call, ...`) for table `t` (`self` is also a keyword recognized by treesitter that would make sense).

The `__index = function(t, k) ... end` meta method here is key and defines how to handle access of keys not found in `t`.

The first segment
```lua
    local raw = rawget(mt, k) -- rawget is table access with bypassing meta stuff
    if raw then
      return raw
    end
```
essentially reroutes to keys defined in `mt`, i.e. `display` and `cwd` (levering closures nicely here for `display`).

```lua
    if k == "Path" then
      t.Path = Path:new(t.value)
      return t.Path
    end

    if k == "path" then
      local retpath = t.value
      if not vim.loop.fs_access(retpath, "R", nil) then
        retpath = t.value
      end
      return t.value
    end
    if k == "stat" then
      local stat = vim.loop.fs_stat(t.value)
      t.stat = vim.F.if_nil(stat, false)
      return stat
    end
```

lazily (i.e., create only on access) defining these attributes.

```lua
    return rawget(t, rawget({ value = 1 }, k))
```
telescope convention that `entry.value` and `entry[1]` will always give you the original "value" as returned by the provider (eg absolute paths from `fd` or tables with lnum col etc from lsp stuff).

Furthermore, compare this to https://github.com/nvim-telescope/telescope.nvim/blob/7bc8efd77bc99f8e0bdf49b68f11ba9de731b45c/lua/telescope/pickers.lua#L43-L44 
where we `__index` is a bit simpler and solely reroutes to the `Picker:$method` which is something you'd see more commonly.

lua/telescope/_extensions/file_browser/make_entry.lua
```diff
@@ -185,6 +188,21 @@ local make_entry = function(opts)
...
+        -- Kinda ick but since stat is a table of tables (and stuff),
+        -- we have to account for those subtables
+        log.warn("Unable to get stat for " .. t.Path:absolute())
```
fdschmidt93, line 194: Just as a side note `t.value` should always be the absolute path already, so should not be needed to infer via `Path`.

Makes me think maybe I should get rid of `Path` where possible. Let's see, quite some work for probably unnoticably better latency :sweat_smile:

lua/telescope/_extensions/file_browser/make_entry.lua
miversen33: I'm not super fond of it as it gives no real feedback to the user, but I completely agree that my solution is a bit big. 

I suppose though that this route doesn't really give the user much for discernable feedback either (they would have to know WHY the date is kicking back as NA which isn't terribly helpful), so maybe going with the simpler approach is the better option. I'm mildly annoyed that I can't figure out a better way to do it with metatables

lua/telescope/_extensions/file_browser/make_entry.lua
miversen33: I think removing path will actually help out quite a bit with performance, plenary does alot under the hood here. 

I tend to see a minimum of 3 duplicate calls to a single libuv function because of it, and plenary calls several with each time they do something. 

That said, that is likely a separate work effort (if you want, feel free to open a new issue and assign it to me)

fdschmidt93: > I think removing path will actually help out quite a bit with performance

Improve performance with regards to what? At most, it should increase latency for up to 1000 initially visible entries (gut feeling tells me 5-10ms) as otherwise `Path` creation is fully deferred and should be appropriately cached within a single instance. Of course it can be improved, but the improvement would be unnoticeable in almost all usage scenarios.

lua/telescope/_extensions/file_browser/make_entry.lua
fdschmidt93: > I'm not super fond of it as it gives no real feedback to the user, but I completely agree that my solution is a bit big.

What kind of feedback would you like to give the user? Viable options I see -

1. Red highlighting as per my screenshot + clarifying in log not stat available - I personally like that one the most
2. `vim.notify` that no stat available -- off of the top of my head, I think is best implemented with a completion callback that clears itself after the first initial loop as in entry maker we don't have info about other entries 
3. Modify `entry.display` to e.g. show `f3 (no stat)` (possibly with 1.) - it's a bit unclear what "no stat" practically means (ie why is no stat available, which actually matters I'd say)

Conclusively, that's why I think the screenshot balances feedback with visual signal, as it's not perfectly clear why exactly we won't have an `stat` I think. 

> When I added this code snippet, I did not get any useful feedback on the UI. I toyed around a bit with it but couldn't get it to give me anything different than if I didn't include it. This is likely just my lack of knowledge of how Telescope works however

Did you set `t.stat = false` if `stat` is `nil` as per diff of my PR?

lua/telescope/_extensions/file_browser/make_entry.lua
miversen33: I'll have to poke this a bit more this weekend, but I noticed that `Path` is a bit slow due to all the wrapping around it. Latency (response time I suppose?) will improve (decrease) as you suggested.
However `telescope-file-browser` is clever and caches (smart!) to help avoid this. Still, I also wonder if removing `Path` makes sense. That said, `telescope-file-browser` uses it for a decent amount of filesystem interfacing _and_ to do logic (even via assumption), so I don't think its a small work effort. Maybe something to explore after a good profiling of the plugin? It might not be the best place to target for optimization (work vs reward)

fdschmidt93: Huh, looks like I've got it right haha

```lua
local bench = require "plenary.benchmark"
local Path = require "plenary.path"

local filename = vim.api.nvim_buf_get_name(0)

bench("Path", {
  warmup = 10,
  runs = 50,
  fun = {
    {
      "Path creation",
      function()
        for _ = 1, 1000 do
          Path:new(filename)
        end
      end,
    },
  },
})

```

Benchmark Group: 'Path' -----------------------                                                                                                                                                 
Benchmark #1: 'Path creation'                                                                                                                                                                   
  Time(mean ± σ):      9.4 ms ±   2.8 ms                                                                                                                                                        
  Range(min … max):    7.7 ms …  18.1 ms  50 runs

Which is practically speaking the maximum latency that can be incurred from Path creation during finding as of now.

cc @Conni2461 another good usage of `plenary.benchmark` :)

lua/telescope/_extensions/file_browser/make_entry.lua
miversen33: > What kind of feedback would you like to give the user? Viable options I see -

**Red highlighting as per my screenshot + clarifying in log not stat available - I personally like that one the most**

I was attempting to get this but I could not until you dropped this little knowledge bomb on me
> Did you set t.stat = false if stat is nil as per diff of my PR? If it's still a table, they evaluate boolean to true, so it thinks there is a stat.

I was _not_ setting to false, I assumed an empty table failed a truthy check. Setting to `false` indeed makes the red highlighting work which is plenty of indication to the end user to me.

Also logs are good :)

I've updated the PR. Thanks for working with me on this!

fdschmidt93: > I was attempting to get this but I could not until you dropped this little knowledge bomb on me

:laughing: 

> Thanks for working with me on this!

Sure thing :)

Final thing I think we have to edit. [telescope.log](https://github.com/nvim-telescope/telescope.nvim/blob/1a91238a6ad7a1f77fabdbee7225692a04b20d48/lua/telescope/log.lua#L3) only logs from `log.info(...)`. With `log.debug` as is, no message appears and no line is written to (for me) `~/.cache/nvim/telescope.log`. Once you set it to `info` or above, you'll get a message and the log entry. I guess `warn` would be the appropriate level?

miversen33: > Final thing I think we have to edit. [telescope.log](https://github.com/nvim-telescope/telescope.nvim/blob/1a91238a6ad7a1f77fabdbee7225692a04b20d48/lua/telescope/log.lua#L3) only logs from log.info(...). With log.debug as is, no message appears (while it should I think) and no line is written to (for me) ~/.cache/nvim/telescope.log. Once you set it to info or above, you'll get a message and the log entry. I guess warn would be the appropriate level?

Ahh ya I forgot to update that, I have all my various log levels set to debug currently, that is a good call. 

Updated the PR to reflect level warn (which I agree is probably the correct level)

fdschmidt93 (approved):
LGTM, thanks :) 

Thank you for your patience on this!

PR #144, opened 2022-05-27
hijack netrw - wipe the buffer matching the directory name

ranebrown:
Prevents extra buffers being left that have the name of an opened directory.

@jamestrew


jamestrew:
good catch! :+1:

fdschmidt93 (approved):
LGTM thanks :)

PR #155, opened 2022-06-30
feat: prompt title accepts string or function

protiumx:
## Motivation
When I'm navigating in different folders I want to see the name of the current folder listed in the finder's prompt title.

## Implementation
Expose `prompt_title` option as function or string.
If a function is passed, it will be evaluated with `current finder` as parameter.
If a string is passed, its value is set to the finder.
If not defined, the default value is a function that will output a string with the current path relative to the `cwd` of the finder.

A `opts.prompt_title_fn` field is added to store the function. This is due to telescope/plenary expecting a string in the `prompt_title` option

## Preview
Command `Telescope file_browser previewer=false theme=dropdown follow=true path=%:p:h`
image

[Kapture 2022-06-30 at 15.41.50.webm](https://user-images.githubusercontent.com/43855513/176692519-e0a88c10-415e-486d-a6f6-2f82ce9e4914.webm)

### DIFF
--- a/lua/telescope/_extensions/file_browser/picker.lua
+++ b/lua/telescope/_extensions/file_browser/picker.lua
@@ -65,11 +65,13 @@ local fb_picker = {}
 ---@field browse_files function: custom override for the file browser (default: |fb_finders.browse_files|)
 ---@field browse_folders function: custom override for the folder browser (default: |fb_finders.browse_folders|)
 ---@field hide_parent_dir boolean: hide `../` in the file browser (default: false)
+---@field collapse_dirs boolean: skip dirs w/ only single (possibly hidden) sub-dir in file_browser (default: false)
 ---@field quiet boolean: surpress any notification from file_brower actions (default: false)
 ---@field dir_icon string: change the icon for a directory (default: )
 ---@field dir_icon_hl string: change the highlight group of dir icon (default: "Default")
 ---@field display_stat boolean|table: ordered stat; see above notes, (default: `{ date = true, size = true }`)
 ---@field hijack_netrw boolean: use telescope file browser when opening directory paths; must be set on `setup` (default: false)
+---@field prompt_title string|function: set prompt title (defaul: function that prints the current path relative to cwd)
 fb_picker.file_browser = function(opts)
   opts = opts or {}
 
@@ -83,26 +85,10 @@ fb_picker.file_browser = function(opts)
   opts.hide_parent_dir = vim.F.if_nil(opts.hide_parent_dir, false)
   opts.select_buffer = vim.F.if_nil(opts.select_buffer, false)
   opts.display_stat = vim.F.if_nil(opts.display_stat, { date = true, size = true })
-  opts.prompt_title = opts.prompt_title ~= nil
 
   local select_buffer = opts.select_buffer and opts.files
   -- handle case that current buffer is a hidden file
   opts.hidden = (select_buffer and vim.fn.expand("%:p:t"):sub(1, 1) == ".") and true or opts.hidden
-  opts.finder = fb_finder.finder(opts)
-  -- find index of current buffer in the results
-  if select_buffer then
-    local buf_name = vim.api.nvim_buf_get_name(0)
-    fb_utils.selection_callback(opts, buf_name)
-    -- opts._completion_callbacks = vim.F.if_nil(opts._completion_callbacks, {})
-    -- table.insert(opts._completion_callbacks, function(current_picker)
-    --   local finder = current_picker.finder
-    --   local selection_index = fb_utils._get_selection_index(buf_name, finder.path, finder.results)
-    --   if selection_index ~= 1 then
-    --     current_picker:set_selection(current_picker:get_row(selection_index))
-    --   end
-    --   table.remove(current_picker._completion_callbacks)
-    -- end)
-  end
 
   function prompt_title_fn(finder)
     local Path = require("plenary.path")
@@ -124,6 +110,23 @@ fb_picker.file_browser = function(opts)
    elseif type(opts.prompt_title) == "string" then
      opts.prompt_title_fn = nil
    end
+
+  opts.finder = fb_finder.finder(opts)
+
+  -- find index of current buffer in the results
+  if select_buffer then
+    local buf_name = vim.api.nvim_buf_get_name(0)
+    fb_utils.selection_callback(opts, buf_name)
+    -- opts._completion_callbacks = vim.F.if_nil(opts._completion_callbacks, {})
+    -- table.insert(opts._completion_callbacks, function(current_picker)
+    --   local finder = current_picker.finder
+    --   local selection_index = fb_utils._get_selection_index(buf_name, finder.path, finder.results)
+    --   if selection_index ~= 1 then
+    --     current_picker:set_selection(current_picker:get_row(selection_index))
+    --   end
+    --   table.remove(current_picker._completion_callbacks)
+    -- end)
+  end
 
   pickers.new(opts, {
     prompt_title = opts.prompt_title,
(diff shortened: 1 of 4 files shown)


fdschmidt93:
Thanks for your PR!

Looks like a cool feature at first glance. I'll review your PR hopefully some time next week, as I'm travelling for the remainder of this one.

protiumx:
Sounds good to me. So we have the `follow` as default be behaviour unless the user defines the title?
I'll add some changes later!

fdschmidt93:
No, default should not change per se.

> Default of that function is what's currently in fb_utils.redraw_border_title

Was meant as a reference to master. I'm unfortunately only on mobile until mid next week. If I recall correctly (apologies if that's wrong), `follow` is mostly useful when only the prompt title is shown (like for dropdown theme) and utility is otherwise largely void.

In other words, titles should be `string|function` such that `follow` would primarily be one function example to implement (and be referred to on the wiki).

Hope that makes sense.

musjj:
For those not patient enough to wait for this PR to get merged, here's a small hack that you can use right now in your config:
```lua
local file_browser = telescope.extensions.file_browser
local fb_utils = require("telescope._extensions.file_browser.utils")
local redraw_border_title = fb_utils.redraw_border_title 
fb_utils.redraw_border_title = function(current_picker)
  local Path = require("plenary.path")
  if current_picker.prompt_border then
    local parent = Path:new(finder.cwd):parent().filename
    local new_title = Path:new(finder.path):make_relative(parent)
    if parent == finder.path then
      new_title = parent
    end
    current_picker.prompt_border:change_title(new_title)
  end
  redraw_border_title(current_picker)
end

local picker = file_browser.file_browser
file_browser.file_browser = function(opts)
  local Path = require("plenary.path")
  local parent = Path:new(opts.cwd):parent().filename
  opts.prompt_title = Path:new(opts.path):make_relative(parent)
  picker(opts)
end
```

protiumx:
@fdschmidt93 sorry for the long time with this PR!
I have made the changes but for some reason I don't see the `redraw_border_title` being called when changing folders in the finder, what am I missing to trigger it so the function can be evaluated again?

lua/telescope/_extensions/file_browser/picker.lua
```diff
@@ -65,7 +65,6 @@ local fb_picker = {}
 ---@field browse_files function: custom override for the file browser (default: |fb_finders.browse_files|)
 ---@field browse_folders function: custom override for the folder browser (default: |fb_finders.browse_folders|)
 ---@field hide_parent_dir boolean: hide `../` in the file browser (default: false)
----@field collapse_dirs boolean: skip dirs w/ only single (possibly hidden) sub-dir in file_browser (default: false)
```
fdschmidt93, line 68: I think line erroneously deleted.

fdschmidt93 (commented):
You need to pass the function to the finder in `fb_finders.finder` in `finders.lua`, as it currently never gets actually passed to the created finder but stays in `opts`.

protiumx:
It's fixed. Let me know if we need some rephrasing for the docs or updates in any other file

musjj:
By the way, is there a way to query the width of the title so that we can trim the title optimally?
Plenary trims the end of the path like this: `long/paths/gets/tri...` by default, which makes it not very useful for longer paths.

fdschmidt93:
@protiumx sorry for not having had a look at this more in depth, on a high level, the PR is fine, I'd probably eventually do some clean up for personal nits I'd have. Nothing to worry about. I'm currently more focused on #210.


> By the way, is there a way to query the width of the title so that we can trim the title optimally? Plenary trims the end of the path like this: long/paths/gets/tri... by default, which makes it not very useful for longer paths.

You can always get the prompt buf with the `get_prompt_buf` util in `make_entry.lua`. Then,

```lua
  local prompt_border_win = require "telescope.state".get_status(prompt_bufnr).prompt_border_win
  local prompt_win_width = vim.api.nvim_win_get_width(prompt_border_win) 
```

and use `align_str` I think from `plenary.strings` to left-truncate the title.

musjj:
Ah, to clarify, I meant to trim it by creation. Since the window isn't created yet, there's no way to check its width.
Maybe the title can be set somewhere after creation?

fdschmidt93:
The PR would have to support that then I guess. You could always embed a defer call or something I guess to recall setting the title for the time being or something along these lines. For the time being, I'd refer to OP though.

protiumx:
> I'd probably eventually do some clean up for personal nits I'd have
If you give me some directions I could implement them. I'm not very happy with this code https://github.com/nvim-telescope/telescope-file-browser.nvim/pull/155/files#diff-57d81f1b1356358e13623948fb3fc7825d9d650aa7c776933e47cfa0bb9cb7b1R104-R112

Perhaps it makes sense we always have a function. If user pass a function, we use that. If user passes a string, we wrap the string in a function. Else we default to the `prompt_title_fn`.

> Ah, to clarify, I meant to trim it by creation. Since the window isn't created yet, there's no way to check its width.
Maybe the title can be set somewhere after creation?

It would be neat to add this to the default Fn but I don't know what would be the best approach. I think it would be better if telescope itself does a left trim or provides an option to trim it since every other extension would have the same problem.
I'd prefer we close this with the working feature and think about trimming in another PR or request a feature in telescope

jamestrew:
This looks to be changing the current default behavior of the prompt title from what I can tell ('File Browser' and 'Folder Browser' depending on which finder you're using).

This new option seems reasonable but I'd like to preserve the current default behavior otherwise it is a semi-breaking feature. 
Depending on the theme you use, having the path in the prompt title is redundant as the result title can already display the current path. 
image

protiumx:
Based on @jamestrew comment I'd suggest this PR adds support for `prompt_title: string|function`
And provide the `promp_title_fn` as an example in the docs. Thoughts?

jamestrew:
I think that could work. I am however wondering how much value exposing the `prompt_title` option as something which can take a function would bring. 
Maybe all we need is a new option something like `prompt_title_show_path (default: false)`. I think this would simplify things for users if they do want the path in the prompt title and they wouldn't have to dig too much into docs or write their own functions.

I don't know, what do you think?
Maybe you can convince me of other use cases but I think 99% of people would just want either:

- "File/Folder Browser" depending on current finder (current default behavior)
- some static string (supported currently with the default telescope `prompt_title` option)
- current path (what this PR would enable)

And having a boolean option would just be easier for users and we'd be able to largely reuse the code for the `results_title`.

protiumx:
Sounds good. I've been using this extension for quite a while now and didn't come up with other use for the prompt title, I just find it useful to always show the full relative path of the current folder, as initially planned on this PR (a `follow` option).
So to sum this up:
- I will add `prompt_title_show_path (default: false)`. It could also be `show_path` imo
- When enabled the prompt title will show the current path of the file browser
If agreed I'll add the changes during the week.
I only need some guidance to deal with truncating the path to fit the window. I guess I can take a look at the implementation of this https://github.com/nvim-telescope/telescope.nvim/blob/master/doc/telescope.txt#L262

jamestrew:
Sounds good.
I think `show_path` isn't specific enough since the current behavior already shows the path in `results_title` on most themes/layouts.

As for the truncating of the path... I kinda don't want to worry about it :sweat_smile:
But if it's frequently cutting off paths for you then yeah maybe making the option something like `prompt_title_path_display table: display current path in prompt title using telescope.defaults.path_display options (default: {})`

You can then do something like
```lua
require'telescope.utils'.transform_path({path_display = opts.prompt_title_path_display}, path)
```

protiumx:
Closing this, it's been a while and I ended up dropping the feature from my personal config as I don't set a prompt title in all the pickers I use

PR #156, opened 2022-07-04
Update README.md

qgadrian:
For deletion, the correct `fb_actions` is `remove` and not `delete`.

This PR updates the README documentation to avoid confusion, but I can see a rename of the internal API to use `delete` instead.


fdschmidt93:
Thanks :)

PR #159, opened 2022-07-07
feat(file_browser): collapse dirs

### LINKED ISSUE #158: "Collapse" directories when a parent directory only contains one child directory (similar to github filebrowser)
I use telescope-file-browser to browse my dotfiles which I manage with gnu/stow. This makes for a big chain of directories which contain only one directory, and it's a bit unefficient to have a picker to select only one option.

The perfect solution would be a 'collapse_directories' option to the already existing list of options to pass to the file_browser picker.

I have considered as an alternative reversing the sorting so that '..' appears last and the cursor is on the folder, which makes you just tap <CR> multiple times until you've reached the last folder in the chain. It works but is far from ideal and prone to error.

fdschmidt93:
With { collapse_dirs = true }
going from (example from stow)
`dotfiles/`
to
`dotfiles/nvim/.config/nvim/`
only requires selecting nvim
in `file_browser` as intermediate
single-folder levels are skipped.
This also works in the reverse
direction.

Closes #158 

@cc paul-ohl

Could you please test this PR with your config? :) just making sure it doesn't result in unintended infinite loops for someone other than me :laughing:

### DIFF
--- a/doc/telescope-file-browser.txt
+++ b/doc/telescope-file-browser.txt
@@ -131,6 +131,9 @@ fb_picker.file_browser({opts})  *telescope-file-browser.picker.file_browser()*
                                              |fb_finders.browse_folders|)
         {hide_parent_dir}   (boolean)        hide `../` in the file browser
                                              (default: false)
+        {collapse_dirs}     (boolean)        skip dirs w/ only single
+                                             (possibly hidden) sub-dir in
+                                             file_browser (default: false)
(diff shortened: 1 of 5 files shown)


paul-ohl:
It works amazing!
No issue found so far.

paul-ohl:
Thank you so much, it's great!

fdschmidt93:
Thanks :)

Issue #152, opened 2022-06-11
[docs] double "from"


LamprosPitsillos:
https://github.com/nvim-telescope/telescope-file-browser.nvim/blob/61a5521853ac739ae3d0d395f52c1178108e0114/doc/telescope-file-browser.txt#L99

PR #160, opened 2022-07-14
fix: typo in documentation

gutyina70:
closes https://github.com/nvim-telescope/telescope-file-browser.nvim/issues/152


fdschmidt93:
Thanks :) but I'm afraid this would have to be changed in the `picker.lua` file from which the docs are generated.

fdschmidt93 (approved):
Thanks :)

PR #171, opened 2022-08-18
Enable to change icon width

delphinus:
Neovim now has `setcellwidths()` ([doc][]) and we can set arbitrary codepoint to `1` or `2` widths despite of the `'ambiwidth'` setting.

[doc]: https://neovim.io/doc/user/builtin.html#setcellwidths()

I set devicon's glyphs can have `2` widths and found all icons become `…` characters. This patch enables to change this for such cases.

| before | after |
|--|--|
| スクリーンショット 0004-08-18 14 50 06 | スクリーンショット 0004-08-18 14 51 19 |

### DIFF
--- a/doc/telescope-file-browser.txt
+++ b/doc/telescope-file-browser.txt
@@ -141,8 +141,6 @@ fb_picker.file_browser({opts})  *telescope-file-browser.picker.file_browser()*
                                              (default: )
         {dir_icon_hl}       (string)         change the highlight group of dir
                                              icon (default: "Default")
-        {icon_width}        (number)         change the width of icons
-                                             (default: 1)
         {display_stat}      (boolean|table)  ordered stat; see above notes,
                                              (default: `{ date = true, size =
                                              true }`)
@@ -435,8 +433,6 @@ fb_finders.finder({opts})            *telescope-file-browser.finders.finder()*
                                        (default: )
         {dir_icon_hl}       (string)   change the highlight group of dir icon
                                        (default: "Default")
-        {icon_width}        (number)   change the width of icons
-                                       (default: 1)
 
 
 
--- a/doc/telescope-file-browser.txt
+++ b/doc/telescope-file-browser.txt
@@ -141,6 +141,8 @@ fb_picker.file_browser({opts})  *telescope-file-browser.picker.file_browser()*
                                              (default: )
         {dir_icon_hl}       (string)         change the highlight group of dir
                                              icon (default: "Default")
+        {icon_width}        (number)         change the width of icons
+                                             (default: 1)
         {display_stat}      (boolean|table)  ordered stat; see above notes,
                                              (default: `{ date = true, size =
                                              true }`)
@@ -433,6 +435,8 @@ fb_finders.finder({opts})            *telescope-file-browser.finders.finder()*
                                        (default: )
         {dir_icon_hl}       (string)   change the highlight group of dir icon
                                        (default: "Default")
+        {icon_width}        (number)   change the width of icons
+                                       (default: 1)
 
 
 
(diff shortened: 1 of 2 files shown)


fdschmidt93:
> I set devicon's glyphs can have 2 widths and found all icons become … characters. This patch enables to change this for such cases.

Do you set that explicitly? I recall that width of glyphs may be terminal emulator dependent (I don't find it super fast, but kitty changes width depending on suffix space of glyph IIRC). Apologies I'm also uninformed on `ambiwidth` and `setcellwidth` because haven't used them and not clear to me what value they add for devicons.

I've also checked upstream telescope and there we concatenate the filename and devicon into one string. I suppose this wasn't done in the original file-browser code I carved the extension out of due to the folder icon. See

https://github.com/nvim-telescope/telescope.nvim/blob/28dc08f614f45d37ad90f170935f1f4e12559aeb/lua/telescope/utils.lua#L435-L463

All in all, I feel this should not be configurable and be handled by `telescope-file-browser` internally just like telescope, but for that we'd kind of need a custom variant of the above linked function, if I see that correctly.

Would you be up to make a PR for that?

delphinus:
> Do you set that explicitly? I recall that width of glyphs may be terminal emulator dependent

Yes, I call `setcellwidths()` to set the widths of icons. Now Neovim in HEAD determines widths of East Asian WIdth Class Ambigous characters with process below.

1. If it find an entry from `setcellwidths()`, use it.
2. Use 1 or 2 width from the `'ambiwidth'` setting.

Surely many terminals (including kitty) have settings for this, but the process above takes priority over them.

Many “Ambiguous” characters, such as box-drawing character, should have single width to draw a valid UI with plugins (including telescope). But I want to draw devicons' emoji characters as double widths in such case.

----

I have checked the implementation and the usage of `utils.get_devicons` and `utils.transform_devicons`.

<dl>
<dt><code>utils.get_devicons</code></dt>
<dd>
<ul>
<li>This returns an icon and its highlight.
<li>This is used <a href="https://github.com/nvim-telescope/telescope.nvim/blob/28dc08f614f45d37ad90f170935f1f4e12559aeb/lua/telescope/make_entry.lua#L575-L575">here</a> and it uses <code>strings.strdisplaywidth</code> to determine the width.
<li>This is for showing rows containing several items with an icon.
</ul>
</dd>
<dt><code>utils.transform_devicons</code></dt>
<dd>
<ul>
<li>This returns a string including an icon and a filename, and a highlight of the icon.
<li>This func supposes that all icons have single width.
<li>This is used <a href="https://github.com/nvim-telescope/telescope.nvim/blob/28dc08f614f45d37ad90f170935f1f4e12559aeb/lua/telescope/make_entry.lua#L310-L310">here</a> and <a href="https://github.com/nvim-telescope/telescope.nvim/blob/28dc08f614f45d37ad90f170935f1f4e12559aeb/lua/telescope/make_entry.lua#L164-L164">here</a>. They also suppose icons have single width.
<li>This is for showing rows containing an icon and a filename only.
</ul>
</dd>
</dl>

For telescope-file-browser, it should show rows containing an icon, a filename and some stats. So we should use `utils.get_devicons`, I think.

I have reverted commits and add one to use `strings.strdisplaywidth` to determine the width. I will rebase if you want only the latest commit for this PR.

(By the way, `utils.transform_devicons` should be fixed to use `strings.strdisplaywidth`. I will PR to telescope later, maybe.)

fdschmidt93 (approved):
Thanks for the run-down! `strdisplaywidth` seems to be exactly what we need to implement trivially what I had a bit much more complicated in my mind from the functions as I was referring to :)
I trust that it manages to show your icons in double width as you want them to and briefly also made sure in simple usage that it doesn't regress anything on my end.

LGTM, thanks :)

PR #193, opened 2022-11-07
`selection_callback` after going to parent dir

matthewgrossman:


lua/telescope/_extensions/file_browser/actions.lua
```diff
@@ -539,6 +540,7 @@ fb_actions.goto_parent_dir = function(prompt_bufnr, bypass)
 
   finder.path = parent_dir
   fb_utils.redraw_border_title(current_picker)
+  fb_utils.selection_callback(current_picker, current_dir_path:absolute())
```
fdschmidt93, line 543: Isn't `finder.path` already an absolute path? Is this actually required?

fdschmidt93:
> I got something working locally in my own branch in this PR (3-liner, super simple): 

That's the idea of the selection callback util :) glad it was that easy.

> you mean such that we can let users opt in/out of this behavior?

I'm torn. I guess it's fine to have as the new default since I suppose the majority of users would appreciate it and we'd have an opt out just in case.

> Also, is the first action_set.select:replace_if code snippet you posted the preferable place because it'll generically handle selecting the right entry, rather than my solution special-casing goto_parent_dir?

Wouldn't it apply to both? I guess I'd introduce one flag to opt out for both usages.

PR #196, opened 2022-12-03
feat(fb_actions.trash): add trash action

### LINKED ISSUE #169: An option to trash files instead of removing them
I would like to put files in the trash instead of removing them.

A generic way to override the `remove` action (so I can change it to the `trash` command).

A configuration flag.

stelcodes:
This action attempts to find a common trash executable (currently "trash" or "gio" in that order) and performs a blocking system call via vim.fn.system to trash each file or directory selected in the file browser. Like actions.remove, refuses to trash parent directory or current working directory.

**More info:**
I really love this plugin! I'm a big fan of avoiding unlinking and using the system trash as much as possible and I think this action would be a valuable asset to the plugin, especially for users who aren't great at Lua or searching GitHub issues but want to use trash for removal. I saw #169 and adapted the work posted by @paveloom. I see there was some interest in adding it to the plugin if a good PR was written.

I think adding a new action is the most appropriate because it completely avoids accidentally breaking `actions.remove`. To make it even easier for users there could be an extension option called `use_trash_for_removal` or something which could make the default binding for `d` use `actions.trash` instead of `actions.remove`.

The `gio` command is a good backup because many Linux systems already have the `glib2` package installed which provides it.

I tested locally with only `trash` available, only `gio` available, and neither installed. Worked great for me. I'm running Linux so I don't know how effective this plugin would be on Mac and Windows, but if neither of those executables are there it will fail fast and do nothing with a nice warning message. Please feel free to correct me on anything.

Prior art:
https://github.dev/ms-jpq/chadtree/blob/ec3f65d3556d0395d95af9db1f06eed266e00761/chadtree/transitions/delete.py#L89
https://github.dev/nvim-tree/nvim-tree.lua/blob/f8489c992998e1e1b45aec65bdb9615e5cd59a61/lua/nvim-tree/actions/fs/trash.lua#L69

Fixes #169


stelcodes:
Just fixed a small typo in the action docstsring.

stelcodes:
Fixed the style guide linter issues

lua/telescope/_extensions/file_browser/actions.lua
```diff
@@ -482,6 +482,94 @@ fb_actions.remove = function(prompt_bufnr)
...
+  vim.ui.input({ prompt = "Trash selections [y/N]: " }, function(input)
+    vim.cmd [[ redraw ]] -- redraw to clear out vim.ui.prompt to avoid hit-enter prompt
+    if input and input:lower() == "y" then
+      for _, p in ipairs(selections) do
+        local is_dir = p:is_dir()
+        local cmd = nil
+        if trash_cmd == "gio" then
+          cmd = { "gio", "trash", "--", p:absolute() }
+        else
+          cmd = { trash_cmd, "--", p:absolute() }
```
MithicSpirit, line 543: `trash` from [trash-cli](https://github.com/andreafrancia/trash-cli) does not accept `--`, so this makes the command always "fail" (although it does still trash the file).

stelcodes: That was a regression now fixed in the most recent version https://github.com/andreafrancia/trash-cli/issues/303#issuecomment-1613465001

emretuna:
any progress on this one?

gmankab:
this really should be merged

PR #197, opened 2022-12-03
fix(fb_actions.remove): fix incorrect log message prefix

stelcodes:
Should be `actions.remove`, not `actions.move`.


fdschmidt93 (approved):
Thanks :)

PR #199, opened 2022-12-10
fix: pass depth option in browse_folders finder

ammarghaus:


fdschmidt93 (commented):
Generally LGTM barring docs as they currently refer to file tree depth and practically speaking this is a breaking change.

My take is that we should have two-layered config:
- general
- file and folder browser

with latter taking precedence if set. That way, the practically speaking breaking change is handled comparably gracefully and gives users appropriate flexibility. Plus options become cleaner.

Would you be happy to implement such a feature / RFC along these lines? If so, we can then quickly coordinate as to how to that best.

fdschmidt93:
Sorry for getting back to you so late, I have implemented trees and I am currently refactoring the `telescope-file-browser.nvim` massively as part of #210 (PR to be renamed :sweat_smile: )

The functionality of this PR will be seamlessly possible with the RFC in which I essentially also take care of

> My take is that we should have two-layered config: [...]

In any case, thanks a lot for your PR and apologies for a bit more wait :) I have a bit of free time on my hands and do some new year's cleaning.

Closing this therefore in favor of #210

PR #207, opened 2022-12-23
feat(file_browser): automatically increase depth with auto_depth

### LINKED ISSUE #200: Automatically increase depth when searching
I would be great to have a simple folder view when navigating with arrows/enter, but automatically switch a deep search mode when typing in the picker.

Use default depth unless the user types something in the prompt. At that point allow searching for nested files and folders (relative to the currently opened folder). When the user deletes the content of the prompt, revert back to the folder view.

I considered binding a different key for different modes, but I think it would be great to have a single view that does both things. There's no conflict and this would allow for both exploring and finding files very efficiently.

Happy to help with the implementation and/or testing!

fdschmidt93:
@rizo is this what you had in mind? :) testing welcome

Close #200

### DIFF
--- a/doc/telescope-file-browser.txt
+++ b/doc/telescope-file-browser.txt
@@ -96,57 +96,61 @@ fb_picker.file_browser({opts})  *telescope-file-browser.picker.file_browser()*
         {opts} (table)  options to pass to the picker
 
     Fields: ~
-        {path}              (string)         dir to browse files from,
-                                             `vim.fn.expanded` automatically
-                                             (default: vim.loop.cwd())
-        {cwd}               (string)         dir to browse folders from,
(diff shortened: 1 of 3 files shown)


rizo:
Thanks @fdschmidt93! I tried usnig this for a few hours and it feels amazing – you get the benefit of nvim-tree and telescope in a single keybinding.

I’ll let you know if I find any issues. Thanks again!

fdschmidt93:
I'd just merge this and maybe mark it as `experimental`: One thing to think about is whether `auto_depth` should be able to excl. directories, for instance. But these sorts of things make this quite a lot more complicated than I guess it needs to be.

Thanks for suggesting a very cool feature :) One thing I didn't think I'd like to have until I've implemented it.

PR #209, opened 2022-12-25
fix(make_entry): path display truncates directory when `grouped = true`

jamestrew:
Fixes regression bug related to #208.

When `grouped = true`, the entries passed to `make_entry` don't have a os separator at the end.
This was resulting in directories being displayed like so with the last letter being truncated:
image


fdschmidt93:
thanks :)

PR #210, opened 2022-12-25
feat(file_browser): tree view

### LINKED ISSUE #165: Option for Tree view when depth value is set
Have the option to view the directories in a tree instead of having to read the whole path to find something.

filetree_W

Something like this but with all the functionality of this plugin!

fdschmidt93:
PR is just a preview. It works reasonably well up until here, but _lots_ of work still required.

E: This PR is well progressed, but I will only be able to continue towards the end of January, since deadline for job soon takes priority ;) 

image

Closes #165

### DIFF
--- a/lua/telescope/_extensions/file_browser/finders.lua
+++ b/lua/telescope/_extensions/file_browser/finders.lua
@@ -23,57 +23,8 @@ local fb_finders = {}
 
 local has_fd = vim.fn.executable "fd" == 1
 
---- Harmonize fd opts for lua config with plenary.scandir in mind.
---- - Note: see also `man fd`
----@param opts table: the arguments passed to the get_tree function
----@field path string: "--base-directory" to search from
----@field depth number: set "--max-depth" if provided
----@field hidden boolean: show "--hidden" entries
----@field respect_gitignore boolean: respect gitignore
----@field add_dirs boolean: false means "--type=file" to only show files
----@field only_dirs boolean: true means "--type=directory" to only show files
----@field threads number: count of threads on which to run
-fb_finders.fd_args = function(opts)
-  local args = { "--base-directory=" .. opts.path, "--absolute-path", "--path-separator=" .. os_sep }
-  if opts.hidden then
-    table.insert(args, "--hidden")
-  end
-  if opts.respect_gitignore == false then
-    table.insert(args, "--no-ignore-vcs")
-  end
-  assert(
-    not ((opts.add_dirs == false) and (opts.only_dirs == true)),
-    "Cannot set conflicting options for add_dirs and only_dirs!"
-  )
-  if opts.add_dirs == false then
-    table.insert(args, "--type=file")
-  end
-  if opts.only_dirs then
-    table.insert(args, "--type=directory")
-  end
-  if not opts.auto_depth and type(opts.depth) == "number" and opts.depth > 0 then
-    table.insert(args, string.format("--max-depth=%s", opts.depth))
-  end
-  -- fd starts much faster (5ms vs 25ms) on single thread for file-browser repo
-  -- only with reasonably large width of directory tree do multiple threads pay off
-  if not opts.auto_depth and (opts.depth < 5 or opts.threads) then
-    table.insert(args, string.format("-j=%s", vim.F.if_nil(opts.threads, 1)))
-  end
-  return args
-end
-
-fb_finders._prepend_tree = function(trees, opts)
-  local args = fb_finders.fd_args(opts)
-  table.insert(trees, 1, args)
-end
-
-fb_finders._append_tree = function(trees, opts)
-  local args = fb_finders.fd_args(opts)
-  table.insert(trees, args)
-end
-
 fb_finders._remove_tree = function(trees, opts)
-  local args = fb_finders.fd_args(opts)
+  local args = fb_utils.fd_args(opts)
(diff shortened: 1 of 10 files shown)


peacock0803sz:
Hi, I'm so interested in this feature, Can I try it out?

LamprosPitsillos:
> Hi, I'm so interested in this feature, Can I try it out?

Yes , you can , and it's good for the developer to report any bugs you might find
Your plugin manager should support `branches` for plugins, `lazy.nvim` does for sure , and just add the `feat/tree` branch

jamestrew:
I can continue development for this if it'll help. Can't promise it'd get it done any sooner but.. :sweat_smile:

fdschmidt93:
I'd have had maybe some time this week to continue but unfortunately got very sick :/

This PR is actually reasonable close, but needs some cleaning up and finishing one more feature I'd like to have. One larger roadblock is that this will definitely be a breaking change and I'd need to look into how to make a release of the "status quo" telescope-file-browser. Probably like 2-3 evenings for me to get to where I want.

> I can continue development for this if it'll help. Can't promise it'd get it done any sooner but.. 

Very much appreciated, but the tree implementation is somewhat involved considering how telescope works. I understand this is a very desired feature and I want to see this go pass the finish line :)

I'd suggest the following, I'll try to finish this towards around mid-March (~2 weeks), then we have a call for beta testers to jump on the branch and hopefully have this by April (another ~2 weeks).

LamprosPitsillos:
Thanks for your hard work! Don't need to rush anything, personal life takes priority!
@fdschmidt93 hope you get better soon.
Also count me in as a beta tester 😄

fdschmidt93:
I think this is usable again, but I'm pretty sure there is still a large amount of subtle issues / unexpected, undefined behavior.

`telescope-file-browser.nvim` backend is like a new, much more complex plugin to both allow "list" and "tree"-stye browsing.

To try this feature, a good default would be (imo) in `telescope.setup.extensions`
```lua
        file_browser = {
          grouped = true,
          initial_browser = "tree",
           -- auto switch to `telescope.builtin.find_files` style finder if there is a prompt
          auto_depth = true,
          depth = 1,
        },
```

Happy to get some very early feedback.
">" and "<" increase and decrease the general folder levels from within insert mode (to be revisited)

If you have a very complex config, there's good chance stuff might not work. The config / plugin will be getting a massive overhaul with this PR.

fdschmidt93:
Testers welcome.

Suggested config

```lua

        file_browser = {
          grouped = true,
          initial_browser = "tree",
           -- auto switch to `telescope.builtin.find_files` style finder if there is a prompt
          auto_depth = true,
          depth = 1,
        },
```
Notes on UX:
- file actions are relative to the current selection: selected files are copied/moved.. to parent folder of file or file in selection
- entering tokens in prompt switches to `find_files` style finder -- file actions should still work as per above!
- Key bindings: < decreases and > increases general depth in insert mode; left arrow goes to parent and right arrow sets the folder into the selected one (i.e. hiding parent)

It should generally be very performant in all actual use cases, one TODO is maybe a time out (albeit 70K files with depth increasing in linux repo is still acceptable in ~1s)

lua/telescope/_extensions/file_browser/finders.lua
```diff
@@ -159,68 +161,135 @@ end
 ---@field git_status boolean: show the git status of files (default: true)
 fb_finders.finder = function(opts)
   opts = opts or {}
+
+  deprecation_notices(opts)
+
+  local auto_depth_defaults = {
+    path_display = vim.F.if_nil(
+      require("telescope.config").pickers.find_files.path_display,
```
jamestrew, line 169: I think this will error out with default telescope settings like `require('telescope').setup()`.

fdschmidt93: Good catch :)

jamestrew:
> from look and feel, I'll probably want to move git info to the end front (next to fs_stat info)

I've been thinking about this as well a bit. When the working tree is clean, having a bunch of empty space prepending the path looks a little awkward. People thought it was a bug.

fdschmidt93:
I moved it to the very front for now, I think that's a good solution. If there's no git info, then there's no spaces for any file.

LamprosPitsillos:
Amazing to see this feature slowly making it to master slowly! Thank you so much , I started testing it right away!

fdschmidt93:
Reporting bugs here is the correct place.

>  For now it looks like depth stays always at 1 with the recommended settings given above. I also tried setting depth to e.x. 3

I cannot reproduce, seems to work as intended for me. Does pressing ">" in insert mode increase depth for you as intended?

The prompt should also show the current "depth=$DEPTH"

LamprosPitsillos:
https://user-images.githubusercontent.com/61395246/228676658-8d0a9b8f-84aa-4ff1-ba4f-15b11e583746.mp4

fdschmidt93:
I'm afraid I'm off to bed now, though unfortunately it's not clear to me why the tree is not appropriately showing for you. I hope to think of something tomorrow.

CleoMenezesJr:
Amazing. Is it a thing yet?

fdschmidt93:
I'm using it personally as my daily driver.
This branch doesn't get updates from `master` (I typically don't need them).

This probably will never make it into `master

PR #222, opened 2023-01-14
feat(file_browser): show git status

### LINKED ISSUE #214: Show Git status on files and folders
I would like to quickly see paths that have modifications or are newly added so that I quickly see the current state of my work.

Have git status indicators on the paths that are changed, similar to other file-browsers (nvim-tree, nerdtree, etc.)

Use `Telescope git_status`, however, I would like to have a single command to open the file overview and increase information density.

**Additional context**

File status with `Telescope git_status`
Screenshot 2022-12-30 at 15 24 32

jan-xyz:
fixes #214

I tested the new implementation on the linux code base and it was reasonably fast with almost no noticeable lag.

How it currently looks like:
Screenshot 2023-01-14 at 23 01 21
Screenshot 2023-01-14 at 23 01 09


fdschmidt93:
Looks cool :)

Sorry, quite swamped these days as I'll have a work deadline next Saturday, so just quick raw pass.

> fdschmidt93 mentioned in the above issue that this should rather be done in the finders.lua function.

What I meant is that `git status` needs to executed in batches of files as single `plenary.jobs` seem incur too much overhead via

```lua
Job:new({ command = "git", args = { "status", "-s", "--", table.concat(entries, " ") } }):sync()
```
where entries a table of absolute paths as returned by `fd` or `plenary.scandir`.

We likely should do parsing before the finder and pass it to `entry_maker`, simplest though may be slow for large repos (100ms for linux in my tests IIRC), but seems to be acceptable in 99% of cases. I think we should go with this and if really needed can improve later.

https://github.com/nvim-telescope/telescope-file-browser.nvim/blob/304508fb7bea78e3c0eeddd88c4837501e403ae8/lua/telescope/_extensions/file_browser/finders.lua#L70

`data` here is the piped output of `fd` with the given options that is a table of absolute paths, this would essentially go in the `entries` in the function referred to above. Then you hand over the parsed output to the entry maker that then displays info pretty much like you did. A more fancy variant would be also copying the display from `telescope` upstream or so.

lua/telescope/_extensions/file_browser/make_entry.lua
```diff
@@ -157,6 +157,16 @@ local make_entry = function(opts)
       table.insert(widths, { width = strings.strdisplaywidth(icon) })
       table.insert(display_array, { icon, icon_hl })
     end
+
+    local git_status, _ = Job:new({ cwd = opts.cwd, command = "git", args = { "status", "--short", "--", path_display } }):sync()
```
fdschmidt93, line 161: Just a general comment in case you'd contribute more in the future to telescope. `path_display` can change tremendously depending on user config. In context of file-browser `entry.value` denotes absolute paths.

jan-xyz: Ah good to know!

I was struggling with getting the correct git status but in the end it was because I was in the wrong working directory and not because of the path. So the relative path (that I'm using the `display_path` here as) shouldn't be needed anymore.

jan-xyz:
> Sorry, quite swamped these days as I'll have a work deadline next Saturday, so just quick raw pass.

All good, I'm glad you found the time anyways to give feedback on this ❤️


> What I meant is that `git status` needs to executed in batches of files as single `plenary.jobs` seem incur too much overhead via

I guess I experienced that overhead first hand now 😅

> ```lua
> 
> Job:new({ command = "git", args = { "status", "-s", "--", table.concat(entries, " ") } }):sync()
> 
> ```
> 
> where entries a table of absolute paths as returned by `fd` or `plenary.scandir`.
> 
> 
> 
> We likely should do parsing before the finder and pass it to `entry_maker`

So I would parse the git status into a table with the paths as keys and the status as value to make it quickly indexable by the `make_entry` function. How do I hand this data over to the `make_entry` function though?

> A more fancy variant would be also copying the display from `telescope` upstream or so.

All it does is color the output I think. It also uses the "raw" git status symbols. I would probably make the symbols configurable?

lua/telescope/_extensions/file_browser/make_entry.lua
```diff
@@ -218,9 +229,11 @@ local make_entry = function(opts)
...
 
-  return function(absolute_path)
+  return function(entry)
+    local absolute_path = entry[1]
     local e = setmetatable({
       absolute_path,
+      git = entry[2],
       ordinal = (absolute_path == opts.cwd and ".")
         or (absolute_path == parent_dir and ".." or absolute_path:sub(cwd_substr, -1)),
     }, mt)
```
jan-xyz, line 239: This was a massive gotcha for me because the telescope docs say
```
It's best practice to have a value reference to the original entry, that way we will always have access to the complete table in our action.
```
(https://github.com/nvim-telescope/telescope.nvim/blob/master/developers.md#entry-maker)

which is not followed here but instead I assume the `__index` function is somehow overwriting this behaviour.

lua/telescope/_extensions/file_browser/finders.lua
```diff
@@ -158,6 +184,7 @@ fb_finders.finder = function(opts)
     select_buffer = vim.F.if_nil(opts.select_buffer, false),
     hide_parent_dir = vim.F.if_nil(opts.hide_parent_dir, false),
     collapse_dirs = vim.F.if_nil(opts.collapse_dirs, false),
+		git_status = vim.F.if_nil(opts.git_status, false),
     -- ensure we forward make_entry opts adequately
```
jan-xyz, line 188:
```suggestion
    collapse_dirs = vim.F.if_nil(opts.collapse_dirs, false),
    git_status = vim.F.if_nil(opts.git_status, false),
    -- ensure we forward make_entry opts adequately
```

jan-xyz:
I re-wrote it to fetch the status upfront. it doesn't work with directories though anymore, because there won't be a direct match. I will find a workaround for this!

lua/telescope/_extensions/file_browser/make_entry.lua
fdschmidt93: > It's best practice to have a value reference to the original entry, that way we will always have access to the complete table in our action.

Mhm, `value` is the original entry (i.e. absolute path) as produced by whatever comes "into" the finder. At least is how I understand that convention.

The complete table in the example refers to the original entries in that context, i.e.

```lua
      results = {
        { "red", "#ff0000" },
        { "green", "#00ff00" },
        { "blue", "#0000ff" },
      }
```

I don't see why you think that's not followed here.

> the __index function is somehow overwriting this behaviour.

The __index metamethod primarily serves lazy attribute creation (e.g. `plenary.Path`, `stat`) and order for access. Strictly speaking we could allow users to pass their own display function if it was part of entry.

lua/telescope/_extensions/file_browser/make_entry.lua
```diff
@@ -218,9 +229,11 @@ local make_entry = function(opts)
     return rawget(t, rawget({ value = 1 }, k))
   end
 
-  return function(absolute_path)
+  return function(entry)
+    local absolute_path = entry[1]
     local e = setmetatable({
       absolute_path,
+      git = entry[2],
```
fdschmidt93, line 236: Without having looked at the rest of the PR here (for brevity most of all) - I think a cleaner pattern here is to pass a `{ absolute path = git_status }` dictionary table to whenever the `entry_maker` is first instantiated and hand over to the finder.

That way you can just put it into the `__index` method and return from the lookup, if the lookup table exists and a corresponding git status for `entry.value` exists.

Seems cleaner for cases in which git status may or may not exist. For instance, the RFC / feat PR will enable an auto switch to something akin to `telescope.builtin.find_files` which should be fully async and hence not show git status. That's done via intermittently creating a new finder. I hope that makes somewhat sense :sweat_smile:

lua/telescope/_extensions/file_browser/make_entry.lua
jan-xyz: That table going into the finder is in this case a list of all the found files as an absolute path in the cwd and their git status like this:
```lua
results = {
        { "/abs/path1", " M" },
        { "/abs/path2", "??" },
        { "/abs/path3", "  " },
      }
```

However, when trying to access `entry.value` from within the `display` function it is referring to the absolute path and not to `{ "/abs/path1", " M" }` what I would expect based on the description from the telescope docs.

lua/telescope/_extensions/file_browser/make_entry.lua
jan-xyz: Makes sense, that might also make the directory case easier to implement to also mark directories which include changed files.

I am just not 100% sure how I would hand this over. Do you have an example? The `results` table was so far the only option I found to pass something to the `entry_maker`. I guess that also in line with me not fully understanding what this `__index` metamethod does.

jan-xyz: Oh. It's this line:
```lua
  local entry_maker = opts.entry_maker { cwd = opts.path }
```
right? I misread it because I'm not used to the trick in Lua to omitting function brackets when calling a function. I would just change it to something like
```lua
  local entry_maker = opts.entry_maker { cwd = opts.path, git_statuses = git_statuses }
```

fdschmidt93: Just a quick 'exactly' to confirm from mobile :)

jan-xyz:
@fdschmidt93 thanks to your pointers I was now able to get it working and it's actually even reasonably fast on the linux codebase as well!

I also updated the description of the PR to reflect this.

https://user-images.githubusercontent.com/5249233/212558580-17d0253e-859b-4bef-b50c-da8cacc6dfab.mov

At the moment it just takes the same highlighting as the dev icon and if that doesn't exist it probably will cause some problem. How do you feel about using the same highlight groups as the Telescope git_status?

fdschmidt93:
Nice!

> How do you feel about using the same highlight groups as the Telescope git_status?

Yep, sounds good :) Off the top of my head, one thing we can try, which I think neo-tree eg does, is to also highlight the path in the same color code. Might be a stupid idea since I virtually never use `:Telescope git_status`

jan-xyz:
I think I'm happy now 😁

I also refactored the git specifics into a separate module.

Looking forward to your review! ❤️

fdschmidt93:
Cool :) As a heads up, I'll most likely have a more depth look after my deadline, probably the Monday in 8 days.

lua/telescope/_extensions/file_browser/finders.lua
```diff
@@ -77,13 +84,23 @@ fb_finders.browse_files = function(opts)
       respect_gitignore = opts.respect_gitignore,
     })
   end
+
+  local file_statuses = {}
```
fdschmidt93, line 88: Stupid nitpick, but `file_statuses` sounds wrong to me and is not meaningful once from within entry maker any longer.

`file_git_status`? not amazing but indicative of `{ file = git_status }` map

lua/telescope/_extensions/file_browser/make_entry.lua
```diff
@@ -157,6 +159,17 @@ local make_entry = function(opts)
       table.insert(widths, { width = strings.strdisplaywidth(icon) })
       table.insert(display_array, { icon, icon_hl })
     end
+
+    if opts.git_status then
+      if entry.value == parent_dir then
+        table.insert(widths, { width = 1 })
+        table.insert(display_array, { " ", icon_hl })
```
fdschmidt93, line 166: I'm reasonably sure off the top of my head 

```lua
        table.insert(display_array, " ")
```

it's just a tuple if there is an associated highlight group.

But yeah, doesn't matter too much.

lua/telescope/_extensions/file_browser/make_entry.lua
```diff
@@ -191,6 +204,22 @@ local make_entry = function(opts)
       return raw
     end
 
+    if k == "file_status" then
```
fdschmidt93, line 207: Similar comment to above, here `git_status` should suffice because now it's associated with the entry directly.

lua/telescope/_extensions/file_browser/make_entry.lua
```diff
@@ -191,6 +204,22 @@ local make_entry = function(opts)
...
+      local file_statuses = vim.F.if_nil(opts.file_statuses, {})
```
fdschmidt93, line 208: Here, we'd create an empty table whenever there is no git status but it's accessed for a file (not sure when that happens, but a bit of an anti pattern).

Something along these lines should probably be moved to the top of the function (before the closure where the entry maker is returned)

```lua
opts.git_file_status = vim.F.if_nil(opts.git_file_status, {})
```

lua/telescope/_extensions/file_browser/picker.lua
```diff
@@ -73,6 +73,7 @@ local fb_picker = {}
 ---@field display_stat boolean|table: ordered stat; see above notes, (default: `{ date = true, size = true }`)
 ---@field hijack_netrw boolean: use telescope file browser when opening directory paths; must be set on `setup` (default: false)
 ---@field use_fd boolean: use `fd` if available over `plenary.scandir` (default: true)
+---@field git_status boolean: show the git status of files (default: false)
```
fdschmidt93, line 76: I'd say we should probably enable this by default even.

In my RFC PR the auto switcher to `find_files` like mode will turn this off then for speed.

fdschmidt93 (approved):
LGTM thank you very much :)

Just minor nitpicks, sorry about those.

I'll test this once I'm free and then merge.

jamestrew:
I've been on this branch for several days and other than an issue with parsing and handling staged vs unstaged items, things have been working smoothly for me.

And I agree, I think enough people would appreciate this feature to have it on as default.

lua/telescope/_extensions/file_browser/finders.lua
jan-xyz: I like how we think alike :D I had the exact same thought process. `file_git_status` just seems very wordy, I was hoping to find something shorter and crisper. I guess `file_git_status` doesn't at least lose the information that we're dealing with information from git. I will change it.

lua/telescope/_extensions/file_browser/make_entry.lua
jan-xyz: What you're trying to say is that the table isn't required? Sounds great! Should I verify this somewhere first?

lua/telescope/_extensions/file_browser/make_entry.lua
jan-xyz: > not sure when that happens, but a bit of an anti pattern

Initially I was thinking that the async case wouldn't pass in any git statuses, but I changed that to now pass in an empty table there, so this is theoretically not required as it should always be there. I can still make it nil-safe with your suggestion though. What do you prefer? Delete or add your suggestion to the top of the function?

lua/telescope/_extensions/file_browser/picker.lua
jan-xyz: Sweet, I'm always a bit afraid of people starting to complain when their default behaviour changes :D I definitely would want this turned on!

jan-xyz:
> issue with parsing and handling staged vs unstaged items

Can you elaborate that a bit more @jamestrew ? Any way for me to reproduce it so I can add that would be nice!

lua/telescope/_extensions/file_browser/make_entry.lua
jan-xyz: I tried it out and it works!

jamestrew:
Ah sorry I forgot the submit my review :sweat_smile: It looks like you were able to figure out most of it from my cryptic comment. 
There's still the edge case with staged/unstage hunks in the same file which hopefully you should see now in my review.

lua/telescope/_extensions/file_browser/git.lua
```diff
@@ -0,0 +1,55 @@
...
+--- Returns a map of absolute file path to file status
+---@param output table: lines of the git status output
+---@param cwd string: the path from which the command was triggered
+---@return table: map from absolute file paths to files status
+M.parse_status_output = function(output, cwd)
+  local parsed = {}
+  for _, value in ipairs(output) do
+    local status = string.sub(value, 1, 2)
+    local file = cwd .. os_sep .. string.sub(value, 4, -1)
+    parsed[file] = status
```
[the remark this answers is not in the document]
jan-xyz: > I think we should trim the status here as well as in the git_abbrev table above.

I would like to avoid that to be able to show different icons and highlight groups depending wether the change is staged or not. I feel like this is a shortcoming in most git status features in file browsers and tools out there. Obviously, it makes it simpler to handle the different cases but it also limits the usability of it.

> Another edge case I just came across is when a file has both staged and unstaged hunks.

Ah nice! I haven't thought of that. Now I also understand what `[AMD]`, `[ MTD]` and `[MTARC]` mean in the git status docs. Those are all the possible statuses this status can be combined with. I would suggest that we then parse `X` and `Y` separately and reserve 2 characters for the status, one for the staged changes (X) and one for the unstaged changes (Y). This way also the table gets smaller and parsing becomes simpler. It would mean that we get annotations like 
```
"~~" - changes both staged and upstaged
"+~" - file added and has unstaged changes
"➡~" - file renamed and has unstaged changes
```

The first column I would also suggest to always color wit `TelescopeResultsDiffAdd` and the second color with `TelescopeResultsDiffChange` or `TelescopeResultsDiffDelete` depending on change or deletion.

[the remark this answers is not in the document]
jan-xyz: This is how it would look like with the double status. I decided against the two hl groups because the plugin doesn't support two hl groups within a single item I think, and otherwise it would add an extra gap to it making the git status even wider.
Screenshot 2023-01-22 at 13 41 26
Screenshot 2023-01-22 at 13 41 31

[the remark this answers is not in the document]
jamestrew: I'm getting different highlighting from you. Specifically the renamed/unstaged changes you show in the last screenshot.
image

Regarding the staged vs unstaged double statuses and the highlighting limitations. I don't particularly love the idea of settling with one color for two statuses which may not share colors. My brain is getting mixed signals seeing different colors for the "changed" status for example depending on the staged/unstaged status (as per the my screenshot).
I think I would either show both statuses with their respective highlights, or go back to showing one status - in the priority of staged > unstaged > none. Leaning towards the latter since show both statuses individually adds an awkward looking gap.

fdschmidt93:
The discussion now went a bit more low-level.

Would one of you guys mind giving me a brief rundown what open questions/decisions to be made are? In case there are any.

jan-xyz:
An open question is how we want to display a mixed state in a file with staged and unstaged changes. Currently, I implemented it in a way that it use both to show two symbols. However, that breaks with the highlight groups because showing the "wrong" highlight groups is really unintuitive and I couldn't figure out how to have two highlight groups in the same display item (maybe that's not possible), but showing the two states in different display items adds a space between the items. If we can't figure out a way to display the two states then we should probably decide if we want to still go with displaying both states or rather prioritize one over the other and only display a single state. That be a bummer though, IMO 😕

lua/telescope/_extensions/file_browser/git.lua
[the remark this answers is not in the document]
jan-xyz: > I'm getting different highlighting from you. Specifically the renamed/unstaged changes you show in the last screenshot.

Between my screenshot and the state I pushed I switched from showing the unstaged changes hl group to the staged hl group.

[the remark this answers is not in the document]
jan-xyz: @jamestrew I reverted to prioritise unstaged changes. Does that make it more intuitive? I've been working with it for the past week and find it actually good.

jan-xyz:
I just found https://github.com/supercrabtree/k which is an ls replacement and also shows git status information. They condense both staged and unstaged changes in a single symbol:

jamestrew:
Hmm.. I hate to sound picky but encoding both staged and unstaged info into one character using both symbols and highlighting feels a little overwhelming :confused: 

I was also thinking about other tools which do something similar and I came across a different `ls` replacement called [exa](https://github.com/ogham/exa). They pretty much directly use the results of `git status -s` (technically they don't use git as a dependency but outcome is similar).
image

Maybe we should keep it really simple, drop the highlighting, stick with just symbols so we can clearly display both statuses in one display item.

lua/telescope/_extensions/file_browser/picker.lua
```diff
@@ -89,6 +90,7 @@ fb_picker.file_browser = function(opts)
   opts.custom_prompt_title = opts.prompt_title ~= nil
   opts.custom_results_title = opts.results_title ~= nil
   opts.use_fd = vim.F.if_nil(opts.use_fd, true)
+  opts.git_status = vim.F.if_nil(opts.git_status, true)
```
jamestrew, line 93: Probably want to check `vim.fn.executable "git"` just in case someone doesn't have `git` now that this option will be true by default.

And add `git` as an optional dependency in the readme.

jan-xyz: I added and updated the documentation for this, do you want to give it another look?

jan-xyz:
> Maybe we should keep it really simple, drop the highlighting, stick with just symbols so we can clearly display both statuses in one display item.

I think this works well for simple statuses like `Add` and `Modify` but merge conflicts will look quite overwhelming I think with the different combinations of statuses by `them` and `us`.

jamestrew:
other than the minor styling issue caught by CI and the my comment above (`vim.F.if_nil(opts.git_status, false)`) LGTM :+1:

@fdschmidt93 thoughts on writing a custom `displayer` function to support having different highlights with one entry item (no separator). this will be a separate PR - one that I can take on.

lua/telescope/_extensions/file_browser/finders.lua
```diff
@@ -158,6 +176,7 @@ fb_finders.finder = function(opts)
     select_buffer = vim.F.if_nil(opts.select_buffer, false),
     hide_parent_dir = vim.F.if_nil(opts.hide_parent_dir, false),
     collapse_dirs = vim.F.if_nil(opts.collapse_dirs, false),
+    git_status = vim.F.if_nil(opts.git_status, false),
```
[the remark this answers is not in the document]
jan-xyz: Ah I updated it and also added it to the doc string!

jan-xyz:
Nice! I fixed the remaining issues!

jamestrew:
@jan-xyz great :rocket:  thanks for sticking through the back and forth. I'll continue to see if we can eventually get the individual highlights to work.

jan-xyz:
It was a lot of fun! I learned a lot and both your and @fdschmidt93 's feedback was very valuable! See you next time ❤️

fdschmidt93:
Many thanks @jan-xyz ! Apologies for going a bit AWOL. I was quite worn out after my deadline end of January and then (personal) life took precedence.

Thanks to @jamestrew for jumping in with review and merging :)

jan-xyz:
There was no rush to this! I am happy I could take this one on and it's good to hear you understand your signals and took care of yourself ❤️

PR #231, opened 2023-02-12
feat: add directory hl for directory entries

### LINKED ISSUE #230: Highlight group for directories
I'd like to be able to highlight directories differently from regular files in the results list, making it easier to distinguish between the two.

A `dir_hl` config option, similar to the existing `dir_icon_hl` but applied to the directory name rather than to its icon.

The `dir_icon` allows you to display an icon beside each directory name, but I prefer no icons. I find it clutters the UI too much and leads to inconsistent alignment of directory names and regular file names.

**Additional context**
This first screenshot shows the [Dirvish](https://github.com/justinmk/vim-dirvish) plugin, which provides a highlight group specifically for directories:

This second screenshot shows telescope-file-browser, which does not provide a way to highlight directories differently from regular files:

jamestrew:
Closes #230 

@fdschmidt93 Nsing the `TelescopePreviewDirectory` hl for directory entries.
This will actually bring parity in how directories are displayed between the entries, file preview and even netrw's explorer.

image

I think this change is reasonable enough to make it a default behavior without adding any new options.

### DIFF
--- a/lua/telescope/_extensions/file_browser/make_entry.lua
+++ b/lua/telescope/_extensions/file_browser/make_entry.lua
@@ -162,6 +162,7 @@ local make_entry = function(opts)
     if #path_display > opts.file_width then
       path_display = strings.truncate(path_display, opts.file_width, nil, -1)
     end
+    path_display = is_dir and { path_display, "TelescopePreviewDirectory" } or path_display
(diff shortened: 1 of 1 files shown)


fdschmidt93 (approved):
Nice, LGTM :)

PR #246, opened 2023-03-08
fix(make_entry): adjust stat correctly on VimResized

### LINKED ISSUE #241: UI: fs_stat display unresponsive to window resizes
### Description

The position of the `fs_stat` component of the results don't response well to changes in the window size.

### Neovim version

```markdown
NVIM v0.9.0-dev-970+gf006313e9
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
```

### Operating system and version

Linux archlinux 6.1.12-arch1-1

### Steps to reproduce

1. `nvim -nu min.lua` 
2. `:Telescope file_browser git_status=false` using `git_status=false` just to see the spacing between result entry items better
3. resize the terminal to force neovim and telescope to resize

### Expected behavior

`fs_stat` info stays right aligned with the results windows during resizes to telescope

### Actual behavior

`fs_stat` info maintained spacing with the file path.

From step 2 above:
image

Step 3 after increasing the window size:
image

Step 3 after decreasing window size:
image
File size is actually gone now.

### Minimal config

```Lua
-- DO NOT change the paths and don't remove the colorscheme
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "cache" }) do
	vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
	vim.fn.system({
		"git",
		"clone",
		"--filter=blob:none",
		"https://github.com/folke/lazy.nvim.git",
		lazypath,
	})
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
	-- add any other plugins here
	"nvim-telescope/telescope-file-browser.nvim",
	dependencies = { "nvim-telescope/telescope.nvim", "nvim-lua/plenary.nvim" },
	config = function()
		require("telescope").setup()
		require("telescope").load_extension("file_browser")
	end,
}
require("lazy").setup(plugins, {
	root = root .. "/plugins",
})

vim.cmd.colorscheme("habamax")
-- add anything else here
```

fdschmidt93:
Closes #241 

@jamestrew could you please confirm this fixes the issue for you as well? :)

### DIFF
--- a/lua/telescope/_extensions/file_browser/make_entry.lua
+++ b/lua/telescope/_extensions/file_browser/make_entry.lua
@@ -1,3 +1,4 @@
+local fb_utils = require "telescope._extensions.file_browser.utils"
 local utils = require "telescope.utils"
 local log = require "telescope.log"
 local entry_display = require "telescope.pickers.entry_display"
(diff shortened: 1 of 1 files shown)


jamestrew:
LGTM thanks.

I just noticed though that the display stat [date]-[size] ordering is nondeterministic looping with `pairs`. Can't believe I didn't notice until now.

PR #252, opened 2023-03-22
Add 'path_prompt_prefix' option

CKolkey:
If enabled, will update the prompt with the relative path from the CWD. Defaults to `false`
Screenshot 2023-03-22 at 15 15 01
Screenshot 2023-03-22 at 15 15 14

### DIFF
--- a/doc/telescope-file-browser.txt
+++ b/doc/telescope-file-browser.txt
@@ -158,9 +158,6 @@ fb_picker.file_browser({opts})  *telescope-file-browser.picker.file_browser()*
         {git_status}        (boolean)         show the git status of files
                                               (default: true if `git`
                                               executable can be found)
-        {prompt_path}       (boolean)         Show the current relative path
-                                              from cwd as the prompt prefix.
-                                              (default: false)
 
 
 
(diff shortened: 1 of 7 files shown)


doc/telescope-file-browser.txt
```diff
@@ -158,6 +158,9 @@ fb_picker.file_browser({opts})  *telescope-file-browser.picker.file_browser()*
         {git_status}        (boolean)         show the git status of files
                                               (default: true if `git`
                                               executable can be found)
+        {prompt_path}       (boolean)         Show the current relative path
```
fdschmidt93, line 161: Docs are generated from lua annotation. This needs to go to `picker.lua`.

fdschmidt93 (approved):
LGTM thanks :) briefly tried it and only needs annotations in lua to correctly generate the docs. Probably best to revert the doc change and then annotate in lua so CI correctly does the job.

CKolkey:
Alright, let's see if that does it :)

fdschmidt93:
Thanks :) I'll have to trigger docgen manually on master after merging. It's often not running especially with first time contributors.

CKolkey:
Ahh, got it. Cool, thanks!

PR #253, opened 2023-03-23
Add file-descriptor mode column

CKolkey:
Screenshot 2023-03-23 at 22 14 47
Screenshot 2023-03-23 at 22 18 34
Screenshot 2023-03-23 at 22 18 21

I noticed that the order of the columns was being shuffled every time you restart neovim, so I changed the iterator slightly so the order is consistent every time. 

I also added `entry.lstat` to be able to accurately display if a fd is a symlink or not, since `stat` _follows_ links.

Finally, `@` is appended to flagged fd's.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ require("telescope").setup {
       quiet = false,
       dir_icon = "",
       dir_icon_hl = "Default",
-      display_stat = { date = true, size = true },
+      display_stat = { date = true, size = true, mode = true },
       hijack_netrw = false,
       use_fd = true,
       git_status = true,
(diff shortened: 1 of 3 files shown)


jamestrew:
Nice!

> I noticed that the order of the columns was being shuffled every time you restart neovim, so I changed the iterator slightly so the order is consistent every time.

This was also kinda bothering me but hadn't gotten around to it so nice again.

I'm not familiar with the `@` symbol. I think that's a macOS thing right?

jamestrew:
Also this is another reason it would be nice to have different highlighting for a single entry item. 
image

I also wonder, if it's worth adding the owner/group info as well. It doesn't have to be in this PR but something to consider.

CKolkey:
It would be super cool to color the individual parts like your screenshot - I considered implementing that next but if you're on it already then thats awesome 🤩

fdschmidt93 (approved):
Thank you very much for your contributions!

Quick test looks like it works as intended and, while I agree we should maybe add owner and character highlighting, let's do that in a follow up PR.

PR #254, opened 2023-03-25
Add action "backspace" for insert mode.

CKolkey:
This one is pretty simple, and if I'm being honest, stolen straight from emacs' `dired`:
In insert mode, If the prompt is empty and the user types `<bs>`, go to parent dir. If there _is_ any user input in the prompt, then behave as normal. 

This pairs very nicely with the new path_prompt feature.

https://user-images.githubusercontent.com/7228095/227717450-1113e881-4eca-4a0f-b0e8-1ac2b9f587a9.mov

### DIFF
--- a/lua/telescope/_extensions/file_browser/actions.lua
+++ b/lua/telescope/_extensions/file_browser/actions.lua
@@ -718,5 +718,16 @@ fb_actions.sort_by_date = function(prompt_bufnr)
   end)
 end
 
+--- If the prompt is empty, goes up to parent dir. Otherwise, acts as normal.
+fb_actions.backspace = function(prompt_bufnr)
+  local current_picker = action_state.get_current_picker(prompt_bufnr)
+
+  if current_picker:_get_prompt() == "" then
+    fb_actions.goto_parent_dir(prompt_bufnr, true)
+  else
+    vim.api.nvim_feedkeys(vim.api.nvim_replace_termcodes("<bs>", true, false, true), "tn", true)
+  end
+end
+
 fb_actions = transform_mod(fb_actions)
 return fb_actions
--- a/lua/telescope/_extensions/file_browser/actions.lua
+++ b/lua/telescope/_extensions/file_browser/actions.lua
@@ -719,11 +719,11 @@ fb_actions.sort_by_date = function(prompt_bufnr)
 end
 
 --- If the prompt is empty, goes up to parent dir. Otherwise, acts as normal.
-fb_actions.backspace = function(prompt_bufnr)
+fb_actions.backspace = function(prompt_bufnr, bypass)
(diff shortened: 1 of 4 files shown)


fdschmidt93:
Cool! Thought about this on your earlier PR as well. From quick glance LGTM, thanks :)

Slight note: should we add a selection callback (sorry on mobile, check other actions please) to retain selection on the directory the user is coming from? I think makes for a good default.

lua/telescope/_extensions/file_browser/actions.lua
```diff
@@ -718,5 +718,16 @@ fb_actions.sort_by_date = function(prompt_bufnr)
   end)
 end
 
+--- If the prompt is empty, goes up to parent dir. Otherwise, acts as normal.
+fb_actions.backspace = function(prompt_bufnr)
```
jamestrew, line 722: Can we make this function also take a `bypass` arg so we're not defaulting to `true` for `goto_parent_dir(prompt_bufnr, true` below?

jamestrew:
> Slight note: should we add a selection callback (sorry on mobile, check other actions please) to retain selection on the directory the user is coming from? I think makes for a good default.

I like this idea. Currently `goto_parent_dir` doesn't implement this either so it'd probably make more sense to make that change there, probably in a separate PR. I'll take a look into that after I type this comment out.

jamestrew:
Also, for my educational purposes, I'm wondering about the `nvim_feedkeys`'s `escape_ks` argument. I can't find anything in the docs specifically about `escape_ks`  but under `nvim_feedkeys` it says:

```
This should be false if you already used |nvim_replace_termcodes()|, and true otherwise.
```
This PR's using `escape_ks = true` despite using `nvim_replace_termcodes`. It appears to work just fine but I'm wondering if there's a reasoning behind it.

lua/telescope/_extensions/file_browser/actions.lua
CKolkey: You got it :)

CKolkey:
> Also, for my educational purposes, I'm wondering about the `nvim_feedkeys`'s `escape_ks` argument. I can't find anything in the docs specifically about `escape_ks` but under `nvim_feedkeys` it says:
> 
> ```
> This should be false if you already used |nvim_replace_termcodes()|, and true otherwise.
> ```
> 
> This PR's using `escape_ks = true` despite using `nvim_replace_termcodes`. It appears to work just fine but I'm wondering if there's a reasoning behind it.

So, to be honest, no. I tried a handful of different combinations to get the behaviour working - the trick was the mode flag to make it a non-recursive mapping. I just forgot to flip this back to what the docs advise. 

Just tested and it works exactly the same, so 🤷🏼

jamestrew:
LGTM thanks

PR #255, opened 2023-03-25
feat(actions): goto_parent_dir selection original dir entry

jamestrew:
Using the `goto_parent_dir` action will now set the picker selection to the directory you just left. 

Additionally, I made a slight refactor/fix so that when you create a directory using the `create` action, the selection is set to the newly created directory. Previously, this was only working for created files.

I have tested these changes against all other actions involving files and folders, and everything works fine except for the `copy` action, which is also broken on the master as well I think

### DIFF
--- a/lua/telescope/_extensions/file_browser/actions.lua
+++ b/lua/telescope/_extensions/file_browser/actions.lua
@@ -107,7 +107,6 @@ fb_actions.create = function(prompt_bufnr)
     if file then
       -- values from finder for values don't have trailing os sep for folders
       local path = file:absolute()
-      path = file:is_dir() and path:sub(1, -2) or path
       fb_utils.selection_callback(current_picker, path)
       current_picker:refresh(finder, { reset_prompt = true, multi = current_picker._multi })
     end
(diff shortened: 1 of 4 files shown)


fdschmidt93:
> Additionally, I made a slight refactor/fix so that when you create a directory using the create action, the selection is set to the newly created directory. Previously, this was only working for created files.

I'm pretty sure on the tree browser branch this is fixed with a general purpose version of `trim_right_os_sep`. Ok for now, I'll unify that once I finalize that larger refactor on the tree branch.

I want to cut a release next week with the status quo of `telescope-file-browser.nvim` and add a CI for tagging versions like lazy.nvim.

lua/telescope/_extensions/file_browser/actions.lua
```diff
@@ -531,6 +530,7 @@ fb_actions.goto_parent_dir = function(prompt_bufnr, bypass)
   local current_picker = action_state.get_current_picker(prompt_bufnr)
   local finder = current_picker.finder
   local parent_dir = Path:new(finder.path):parent():absolute()
+  local current_dir = Path:new(finder.path):absolute()
```
fdschmidt93, line 533: `finder.path` should always be absolute. Is there a scenario where this is needed?

jamestrew: No you're right. I'm switched to using `finder.path` directly.

fdschmidt93 (approved):
I'll merge and refactor the path sanitization in the tree branch.

Thanks :)

PR #256, opened 2023-03-25
multi-color entry display

jamestrew:
Tweak the passing of display items to enable individually highlighted entry items.
image

### DIFF
--- a/lua/telescope/_extensions/file_browser/stats.lua
+++ b/lua/telescope/_extensions/file_browser/stats.lua
@@ -1,0 +1,95 @@
+local M = {}
+
+local SIZE_TYPES = { "", "K", "M", "G", "T", "P", "E", "Z" }
+local YEAR = os.date "%Y"
+
+local SIZE_HL = "TelescopePreviewSize"
+M.size = {
+  width = 5,
+  right_justify = true,
+  display = function(entry)
+    local size = entry.stat.size
+    for _, v in ipairs(SIZE_TYPES) do
+      local type_size = math.abs(size)
+      if type_size < 1024.0 then
+        if type_size > 9 then
+          return { string.format("%3d%s", size, v), SIZE_HL }
+        else
+          return { string.format("%3.1f%s", size, v), SIZE_HL }
+        end
+      end
+      size = size / 1024.0
+    end
+    return { string.format("%.1f%s", size, "Y"), SIZE_HL }
+  end,
+}
+
+local DATE_HL = "TelescopePreviewDate"
+M.date = {
+  width = 13,
+  right_justify = true,
+  display = function(entry)
+    local mtime = entry.stat.mtime.sec
+    if YEAR ~= os.date("%Y", mtime) then
+      return { os.date("%b %d  %Y", mtime), DATE_HL }
+    end
+    return { os.date("%b %d %H:%M", mtime), DATE_HL }
+  end,
+}
+
+local color_hash = {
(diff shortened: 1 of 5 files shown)


fdschmidt93:
> This can probably be implemented upstream eventually but I think telescope-file-browser is probably a solid testing ground.

I'd much prefer this to go to `telescope.nvim` directly. The reason is I definitely want not to diverge in any way and the "source of truth" of such functions should be `telescope.nvim`.

> This lets us have different color highlights for git status signs and file permissions.

This is already possible. Its undocumented (see [here](https://github.com/nvim-telescope/telescope.nvim/blob/942fe5faef47b21241e970551eba407bc10d9547/lua/telescope/pickers/entry_display.lua#L112-L115)), but in fact the entries passed to the displayer are

`{ entry: table, hl: string | function }`

where, if a function, an iterator (table, essentially) returns over highlights by string offset.

Here's two examples from my live grep extension, [simple table](https://github.com/fdschmidt93/telescope-egrepify.nvim/blob/607c461d6fad7a157952f4a2a7e82b6d11045561/lua/telescope/_extensions/egrepify/entry_maker.lua#L18-L34), [iterating over submatches in rg json](https://github.com/fdschmidt93/telescope-egrepify.nvim/blob/607c461d6fad7a157952f4a2a7e82b6d11045561/lua/telescope/_extensions/egrepify/entry_maker.lua#L38-L70)

```lua
local displayer = entry_display.create {
  separator = " ",
  items = {
    { width = opts.bufnr_width, separator = " " },
    { width = 4, separator = false }, -- equivalent to `separator = ""`
    { width = icon_width, separator = ">", separator_hl = "CursorLine" },
    { remaining = true },
  },
}
```
The added value appears mostly (subjective) ergonomics, for which I don't want to diverge from upstream though. `{ entry: table, hl: function }` is a clean interface (but of course only once you know about it) for issues your implementation proposes to solve.

I'm very happy to have more granular entry highlighting, but for the time being only with the existing interface.

jamestrew:
I've reworked the PR to use `{ display: string, hl: string | function }`.

> clean interface (but of course only once you know about it) 

Agreed, it was pretty easy to use. I might work on the docs for it upstream if I ever get around to it.

lua/telescope/_extensions/file_browser/stats.lua
```diff
@@ -0,0 +1,95 @@
...
+  right_justify = true,
+  display = function(entry)
+    -- TODO: maybe just iterate over the final string?
+    local owner, group, other = string.format("%3o", entry.stat.mode):match "(.)(.)(.)$"
+    local stat = vim.tbl_flatten {
+      mode_type_map[entry.lstat.type] or "-",
+      mode_perm_map[owner],
+      mode_perm_map[group],
+      mode_perm_map[other],
+      " ",
```
fdschmidt93, line 77: I agree with removing

```lua
entry.stat.flags ~= 0 and "@" or " "
```

since I understand from the original PR it's macOS specific and doesn't add value anyways? emacs' dired is also not showing.

Let's accordingly adjust `width = 10` and remove final `" "` completely so highlighting without the last whitespace is correct (unless there's good reason to keep it).

fdschmidt93 (approved):
Thansk, LGTM and a lot cleaner I think :)

Stupid nit: I like carving out `fs_stat` into a separate file but `stats.lua` is uninformative. Let's just name it `fs_stat.lua` and I think for new people navigating the code base it's a lot clearer what's supposed to be there. Then we can merge.

PR #258, opened 2023-03-27
fix: entry width calculation with separator

jamestrew:
Fixes a small error in the entry width calculation related to the separator width.
This was leading to ever worsening trailing spaces at the end of entries as we add more `display_stat` items.

Before:
image
After (with fix):
image
(btw experimenting with adding owner/group fs_stat info but the width can't be static due to non-deterministic username lengths so not super simple - maybe not worth it)

I've tested this using a mix both options for `git_status` and `disable_devicons` and looks good.

### DIFF
--- a/doc/telescope-file-browser.txt
+++ b/doc/telescope-file-browser.txt
@@ -79,6 +79,8 @@ fb_picker.file_browser({opts})  *telescope-file-browser.picker.file_browser()*
       - `<C-f>/f`: Toggle between file and folder browser
       - `<C-h>/h`: Toggle hidden files/folders
       - `<C-s>/s`: Toggle all entries ignoring `./` and `../`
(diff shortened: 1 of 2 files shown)


fdschmidt93:
> btw experimenting with adding owner/group fs_stat info but the width can't be static due to non-deterministic username lengths so not super simple - maybe not worth it

It's not super simple, but shouldn't be too hard either. It's mostly dependent on the largest file name. In other words, in your screenshots of the top off my head it would just fit without issues.

The problem is more that we'd need to add a ranking and omit fs_stat in some order if the space is too narrow.

fdschmidt93 (approved):
Otherwise LGTM at a glance :) Haven't tested it but diff looks correct and works for you in tests.

PR #271, opened 2023-05-21
fix(action): update multi_icon with select/toggle_all actions

etherswangel:
the '+' icon not updated when select_all/toggle_all

solution found in [telescope.nvim #1682](https://github.com/nvim-telescope/telescope.nvim/pull/1682), works for me.


jamestrew:
LGTM thanks

jamestrew (approved):

Issue #300, opened 2023-08-13
Symlinked folders are not shown


aadi58002:
### Description

I use stow to manage my config file. All the folders which stow has symlinked are not visible inside the directory view.

![image](https://github.com/nvim-telescope/telescope-file-browser.nvim/assets/96418027/30c781a8-099d-4a59-99de-840cdae841aa)

![image](https://github.com/nvim-telescope/telescope-file-browser.nvim/assets/96418027/2fd2a9f3-c325-4c54-bdc7-603460fb274f)

### Neovim version

```markdown
NVIM v0.9.1
Build type: None
LuaJIT 2.1.0-beta3
Compilation: /usr/lib/ccache/bin/cc  -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wshadow -Wconversion -Wvla -Wdouble-promotion -Wmissing-noreturn -Wmissing-format-attribute -Wmissing-prototypes -fno-common -Wimplicit-fallthrough -fdiagnostics-color=always -fstack-protector-strong -DUNIT_TESTING -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -I/usr/include/luajit-2.1 -I/usr/include -I/usr/include -I/builddir/neovim-0.9.1/build/src/nvim/auto -I/builddir/neovim-0.9.1/build/include -I/builddir/neovim-0.9.1/build/cmake.config -I/builddir/neovim-0.9.1/src -I/usr/include -I/usr/include -I/usr/include -I/usr/include -I/usr/include -I/usr/include -I/usr/include

  system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/share/nvim"
```

### Operating system and version

Void Linux

### Steps to reproduce

1. Create a soft symlink to any folder.
2. Open the folder where the symlink was created.
3. The symlink is not visible in the file and folder view.

### Expected behavior

It expected to show symlinks like (Like in the screen shot) with a arrow and path of the link.It can also show them as just regular folders or files. 

### Actual behavior

Symlinks are ignored and not visible in the default view

### Minimal config

```Lua
vim.cmd [[set runtimepath=$VIMRUNTIME]]
vim.cmd [[set packpath=/tmp/nvim/site]]
local package_root = '/tmp/nvim/site/pack'
local install_path = package_root .. '/packer/start/packer.nvim'
local function load_plugins()
  require('packer').startup {
    {
[...]

PR #312, opened 2023-09-17
Add new option `follow-symlinks` to traverse symbolic links

rashil2000:
As per the discussion in #300


doc/telescope-file-browser.txt
jamestrew, line 1: the help docs for telescope are all auto-generated from docstrings so this is not necessary. should be fine though i think.

README.md
```diff
@@ -106,6 +106,7 @@ require("telescope").setup {
       select_buffer = false,
       hidden = { file_browser = false, folder_browser = false },
       -- respect_gitignore
+      -- follow_symlinks
```
jamestrew, line 109: this can just be `follow_symlinks = false` uncommented

jamestrew (commented):
Thanks I think we're looking mostly good here. Just a couple of small tweaks and I'll merge this.

rashil2000:
Done, thanks for the review :)

jamestrew (approved):

PR #317, opened 2023-09-20
Autocreate prompt content if entry is nil

CKolkey:
I found it very annoying to need to hit shift-enter instead of enter. So... Now you just need to hit enter and the prompt will get created.


jamestrew:
I don't think this makes `<s-cr>` redundant does it? You can `create_from_prompt` without an empty entry with `<s-cr>` but switching to `<cr>` would now require the entries to be empty.

CKolkey:
Fair point! I put it back.

jamestrew (approved):

jamestrew:
thanks!

PR #319, opened 2023-10-04
Fix default mappings unmapping

backdround:
There is a problem with unmapping default extension keys.  
Example:
```lua
mappings = {
  i = {
    ["<M-m>"] = false,
  },
},
```
will end up with something like this eventually:
```lua
mappings = {
  i = {
    ["<M-m>"] = false,
    ["<A-m>"] = fb_actions.move,
  },
},
```
I didn't retrieved all the results, that depend on which mapping will be proceeded first, but my experience is that I have my global mapping muted (it doesn't work and the default mapping doesn't work either)

So I've rewritten part with mappings merging. Current algorithm:
- retrieves all ignored mappings (in internal termcodes)
- merges all mappings
- removes all ignored mappings (comparing with internal termcodes)

### DIFF
--- a/lua/telescope/_extensions/file_browser/config.lua
+++ b/lua/telescope/_extensions/file_browser/config.lua
@@ -116,11 +116,52 @@ local hijack_netrw = function()
   })
 end
 
+local unite_mappings = function(...)
+  local add_ignored_keys_from_keymap = function(ignored_keys, mode, keymap)
+    ignored_keys[mode] = ignored_keys[mode] or {}
+    for key, action in pairs(keymap) do
+      if action == false then
+        local internal_key = vim.api.nvim_replace_termcodes(key, true, true, true)
+        ignored_keys[internal_key] = true
+      end
+    end
+  end
+
+  -- Fetch all ignored keys
+  local ignored_keys = {}
+  for _, mappings in ipairs({...}) do
+    for mode, keymap in pairs(mappings) do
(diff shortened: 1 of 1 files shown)


jamestrew:
Sorry for the late response.
I'm not sure how to feel about this. The current behavior seems to be consistent with telescope's builtin pickers.
```lua
require("telescope").setup({
  defaults = {
    mappings = {
      i = {
        ["<A-m>"] = false,
      },
    },
  },
  pickers = {
    find_files = {
      mappings = {
        i = {
          ["<M-m>"] = function() print("hello") end,
        },
      },
    },
  },
})
```
I think if we were to implement a change like this, it should probably be in upstream telescope.
What do you think?

backdround:
I've just tested it with telescope's mappings, and telescope's mappings work as expected with `<A-m>` and `<M-m>` either.

My original problem was that I tried to disable all of the plugin's mappings and I didn't figure out how I can do this.

for `telescope` for example, I can simply use `default_mappings = {}`:
```lua
telescope.setup({
  defaults = {
    default_mappings = {},
  },
})
```

Unfortunately, this plugin doesn't support this approach.

jamestrew:
Ah okay. For disabling default mappings see https://github.com/nvim-telescope/telescope-file-browser.nvim/issues/322#issuecomment-1793981447

backdround:
Thanks. I did it at the time of the pull request.

It it isn't required, I'll close it

PR #323, opened 2023-10-20
Adding the show symlink option to finder

aadi58002:
Addresses the issue #300

### DIFF
--- a/lua/telescope/_extensions/file_browser/actions.lua
+++ b/lua/telescope/_extensions/file_browser/actions.lua
@@ -811,7 +811,7 @@ fb_actions.path_separator = function(prompt_bufnr)
   local current_picker = action_state.get_current_picker(prompt_bufnr)
   local dir = Path:new(current_picker.finder.path .. os_sep .. current_picker:_get_prompt() .. os_sep)
 
-  if dir:exists() and dir:is_dir() then
+  if current_picker.finder.files and dir:exists() and dir:is_dir() then
     fb_actions.open_dir(prompt_bufnr, dir.filename)
   else
     vim.api.nvim_feedkeys(vim.api.nvim_replace_termcodes(os_sep, true, false, true), "tn", false)


aadi58002:
I actually want to implement to show the path of the symlink ex below.

jamestrew:
Sorry for the late response.

We can get the path of the symlink using a libuv function `fs_readlink` https://github.com/luvit/luv/blob/master/docs.md#uvfs_readlinkpath-callback
We can access these libuv function through the `vim.loop` namespace.

image
image


We can make the call to it similar to how we call `vim.loop.fs_stat` here: https://github.com/nvim-telescope/telescope-file-browser.nvim/blob/6e51d0cd6447cf2525412220ff0a2885eef9039c/lua/telescope/_extensions/file_browser/make_entry.lua#L226-L232

Then we should be able to just modify `path_display` if `opts.show_symlink` is true and concatenate `entry.link_path` (or whatever you call the `k == "stat"` equivalent).

aadi58002:
I inspected the entry element in the make_entry file and saw that lstat and the stat values on the entry is always same despite the file being a link or not. I don't think this is a desired behavior (maybe).

```diff
diff --git a/lua/telescope/_extensions/file_browser/make_entry.lua b/lua/telescope/_extensions/file_browser/make_entry.lua
index b8cf898..f99203f 100644
--- a/lua/telescope/_extensions/file_browser/make_entry.lua
+++ b/lua/telescope/_extensions/file_browser/make_entry.lua
@@ -124,6 +124,8 @@ local make_entry = function(opts)
     local display_array = {}
     local icon, icon_hl
     local is_dir = entry.Path:is_dir()
+    
+    print(vim.inspect(entry.lstat))
     -- entry.ordinal is path excl. cwd
     local tail = fb_utils.trim_right_os_sep(entry.ordinal)
     -- path_display plays better with relative paths
```
This line and :messages to inspect the element ( i don't know if there is a better way 😄 )

jamestrew:
I tried putting this in the same line
```lua
    vim.print { path = entry.path, _type = entry.lstat.type }
```
and it's showing `directory`/`file`/`link` for the `lstat.type` accordingly. (I'm also just using `:message` as well).
Do you see this? Make sure you have the `show_symlinks` option enabled.

aadi58002:
image

It does not seem to work for me
image

These directories are most symlinks but it still shows them as directories

jamestrew:
maybe verify the flags being passed to `fd`
All I'm doing is hardcoding this to `fd_file_args`
```lua
  table.insert(args, "--type")
  table.insert(args, "symlink")

```

We actually already have builtin support for showing directory vs link vs files in the fs stats so with the code above I get this
image
Notice the `d` vs `l` in the fs mode/permissions column.

aadi58002:
It seems the `--follow` options from the #312 PR makes the symlinks appear as directories.

jamestrew:
I see two ways the `follow_symlinks` affects file_browser:
1. showing symlinks (traversed) in the result
2. the way the links are opened/cd'ed (as a result of fd returning traversed link path)

I think with the `show_symlinks` option can be mutually exclusive with `follow_symlinks` in terms of fd args. So if `show_symlinks=true` then don't pass `--follow` to fd. I think this makes sense because in terms of showing results, they have a similar purpose. With `show_symlinks` we can have the additional benefit of showing both the link path and true path (`x -> y`).

In terms of cd'ing and opening files, `follow_symlinks` should continue to follow the symlink regardless of `show_symlinks`.

Does that make sense?
Basically if a user have `show_symlinks=true` but `follow_symlinks=false` and they hit `<CR>` on `dir_path/ -> /true/path/to/dir/`, file_browser will cd into `dir_path/` but if `follow_symlinks=true` than it should  cd into `/true/path/to/dir/`.

So you'll  have to adjust which path `entry.path` points to based on `follow_symlinks` here https://github.com/nvim-telescope/telescope-file-browser.nvim/blob/3044ff9e38d1ed8d7818d72d9f951ed9d1b0563d/lua/telescope/_extensions/file_browser/make_entry.lua#L222-L224

^ `t.value` here is the `absolute_path` from index 1 of this metatable
https://github.com/nvim-telescope/telescope-file-browser.nvim/blob/3044ff9e38d1ed8d7818d72d9f951ed9d1b0563d/lua/telescope/_extensions/file_browser/make_entry.lua#L248-L253

aadi58002:
Sorry about this made a little mistake with git and deleted the remote branch i reopened a new PR i also made a pseudo working example which shows the link.

PR #325, opened 2023-11-01
fix: incorrect path separator of "../" on Windows

futsuuu:
use `os_sep`(`require("plenary.path").path.sep`), instead of `/`


jamestrew (approved):
thanks!

PR #328, opened 2023-11-05
feat: add `no_ignore` option for fd

### LINKED ISSUE #326: Custom `fd` args
I have `.ignore` files in some places to prevent regular `fd` searches from having some files show up, but I don't want telescope-fb to ignore these files.

I would like to use the `--no-ignore` flag to tell `fd` to not heed the `.ignore`. Since there are no current options to do this, I believe that the best way would be to add an option for arguments that should be passed to `fd`. This would also overall increase flexibility.

- I could just use a `.gitignore`, but that'd annoy me since this isn't a git repo;
- Changing `respect_gitignore` to use `--no-ignore` (instead of `--no-ignore-vcs`), but then the name isn't accurate and there would be significantly different behavior between fd and other backend(s?);
- Adding a new option for `--no-ignore` (separate from `respect_gitignore`), but then the interaction between this option and `respect_gitignore` is unintuitive for new users.

jamestrew:
Adds a new `no_ignore` option (only applicable to fd) which enables fd's `--no-ignore` option

closes #326

### DIFF
--- a/lua/telescope/_extensions/file_browser/finders.lua
+++ b/lua/telescope/_extensions/file_browser/finders.lua
@@ -62,9 +62,13 @@ local function fd_file_args(opts)
   if not opts.respect_gitignore then
     table.insert(args, "--no-ignore-vcs")
   end
+  if opts.no_ignore then
+    table.insert(args, "--no-ignore")
+  end
   if opts.follow_symlinks then
     table.insert(args, "--follow")
   end
+  vim.print(args)
   return args
 end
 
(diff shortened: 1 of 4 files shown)


lua/telescope/_extensions/file_browser/finders.lua
```diff
@@ -62,9 +62,13 @@ local function fd_file_args(opts)
   if not opts.respect_gitignore then
     table.insert(args, "--no-ignore-vcs")
   end
+  if opts.no_ignore then
+    table.insert(args, "--no-ignore")
+  end
   if opts.follow_symlinks then
     table.insert(args, "--follow")
   end
+  vim.print(args)
```
MithicSpirit, line 71: Debug print left in accidentally? Or is this intended?

jamestrew: that was unintended :grin: 
thanks for taking a look

PR #334, opened 2023-11-13
 Adding the show_symlink option to finder

aadi58002:
Addresses the issue #300 

And also the reopen issue for #323 ( which was closed by mistake )


lua/telescope/_extensions/file_browser/make_entry.lua
```diff
@@ -159,6 +161,10 @@ local make_entry = function(opts)
       end
     end
 
+    if is_symlink then
```
jamestrew, line 164: nit: I think we can just do `entry.lstat.type == "link"` here directly since we only check that once here anyways

lua/telescope/_extensions/file_browser/make_entry.lua
```diff
@@ -238,6 +244,7 @@ local make_entry = function(opts)
         t.lstat = false
       else
         t.lstat = lstat
+        t.readlink = vim.F.if_nil(vim.loop.fs_readlink(t.value),false)
```
jamestrew, line 247: I'm thinking let's move this to it's own conditional block and use `fs_realpath` instead.
```lua
    if k == "realpath" then
      t.realpath = vim.F.if_nil(vim.loop.fs_realpath(t.value))
      return t.realpath
    end
```

lua/telescope/_extensions/file_browser/make_entry.lua
```diff
@@ -159,6 +161,10 @@ local make_entry = function(opts)
       end
     end
 
+    if is_symlink then
+        path_display = string.format("%s -> %s ",path_display,entry.readlink);
```
jamestrew, line 165: We should probably use `transform_path` on the real path as well
```suggestion
        path_display = string.format("%s -> %s", path_display, utils.transform_path(opts, entry.realpath));
```

jamestrew (changes requested):
I think we're on the right track.
Besides the changes I mentioned, we also need to tackle the `<cr>` behavior when `show_symlinks = true` and `follow_symlinks = false`.

Right now, the `open_dir` action already calls `fs_realpath` so it follows symlinks regardless of the `follow_symlinks` option. I think this can be considered a bug.
Probably want to take that out. We can dictate which path (real or symlink) `<cr>` will act on here: https://github.com/nvim-telescope/telescope-file-browser.nvim/blob/d7c453396a043c265bef1227920628e7b811ff30/lua/telescope/_extensions/file_browser/make_entry.lua#L222-L224 by setting this to realpath if `follow_symlinks = true` or keep it `t.value`.

Also CI is failing. I think you just need to run stylua.

doc/telescope-file-browser.txt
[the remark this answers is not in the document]
aadi58002: Didn't knew that 😄

aadi58002:
> I think we're on the right track. Besides the changes I mentioned, we also need to tackle the `<cr>` behavior when `show_symlinks = true` and `follow_symlinks = false`.
> 
> Right now, the `open_dir` action already calls `fs_realpath` so it follows symlinks regardless of the `follow_symlinks` option. I think this can be considered a bug. Probably want to take that out. We can dictate which path (real or symlink) `<cr>` will act on here:
> 
> https://github.com/nvim-telescope/telescope-file-browser.nvim/blob/d7c453396a043c265bef1227920628e7b811ff30/lua/telescope/_extensions/file_browser/make_entry.lua#L222-L224
> by setting this to realpath if `follow_symlinks = true` or keep it `t.value`.
> 
> Also CI is failing. I think you just need to run stylua.

I tried to create a basic implmentation for it but i have a issue

t.symlinks_paths_list is a table with integer indexs which should contain table of format
`{ realpath = "path_symlink_points_to" , path = "path_of_the_symlink" }`

```lua
    if k == "path" then
      local path = t.value
      for index = #t.symlinks_paths_list,1, -1 do
        value = t.symlinks_paths_list[index]
        local temp = string.gsub(path,value.realpath,value.path,1)
        if temp == path then
            table.remove(t.symlinks_paths_list,index)
        else
            path = temp
        end
      end
      return path
    end

```

But i don't know where i can get the list of current symlinks i have enteried 
ex:-
```
/a/b -> d
/d/e -> f
/f
```
If we go from b to d then the symlinks_paths_list should have { 1: {realpath = "/d" , path = "/a/b" }}
After you go to b then it should become { 1: {realpath = "/d" , path = "/a/b"} , 2: {realpath = "/f" , path = "/a/b/e"}}


I can't seem to find a place where to store the symlinks list persistently across sessions of the file-browser

gmankab:
this really should be merged

PR #336, opened 2023-11-15
fix(actions.move): correctly rename open (dir) buffers

fdschmidt93:
Wild that this bug is on `master` since initial commit.
I had this fixed on the `tree` branch (with a slight, unrelated bug).

Problem: open buffers are not correctly renamed when their parent(s) directories or the buffer itself is moved.
Solution: rename moved buffers

Add note on order sensitivity. Maybe we should not allow unpacking, but then we would have to recursively check whether any parent is included.


jamestrew (approved):
lgtm!
yeah I can't believe no one ran into this (and reported it).

PR #342, opened 2023-12-03
fix(action.move): moving folders

jamestrew:
`vim.fs.basename("/foo/bar/)` returns an empty string. This behavior differs from the unix `basename` program which returns `bar`.

Just need to accommodate for this.


che247:
This seems to have broken moving individual files. I noticed that whenever I marked a file to be moved and moved it, it would rename that file to whichever folder name it originally came out from. To confirm these changes caused the issue, I commented lines 350-352 and attempted to move a file again and it worked once again.

jamestrew:
@che247 oh man classic lua truthy-ness issue got me.
thanks for catching that. I've fixed this in the latest commit.

che247:
no problem! thank you for your work in this!

PR #344, opened 2023-12-07
fix(action.move): moving folders

jamestrew:
`vim.fs.basename("/foo/bar/)` returns an empty string. This behavior differs from the unix `basename` program which returns `bar`.

Just need to accommodate for this.


jamestrew:
related #342 #343

PR #360, opened 2024-02-13
feat: mode option for files and directories

LeFrosch:
Hi! While using this plugin I had problems with the default permissions used to create new folders (`plenary.nvim` uses 0700 to create directories). I could not find any option to change the permissions for newly create directories. Therefore, this PR adds the following options:
- `dir_mode`: the default mode/permissions used to create new directories (default to 448 -- decimal for 700)
- `file_mode`: the default mode/permissions used to create new files (defaults to 420 -- decimal for 644)

I took the defaults from the `Path.mkdir` and `Path.touch` methods from `plenary.nvim`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,8 @@ require("telescope").setup {
       hijack_netrw = false,
       use_fd = true,
       git_status = true,
+      dir_mode = 448, -- decimal for 700
+      file_mode = 420, -- decimal for 644
       mappings = {
         ["i"] = {
           ["<A-c>"] = fb_actions.create,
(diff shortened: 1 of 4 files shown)


jamestrew:
Thanks, that's strange that plenary's default is 700. I didn't notice that. I wonder if that's an error.

But I would actually prefer not to have to take on these new options.
Rather I would like to change the default to 755, the default for `mkdir`. I think the default for `Path.touch` matches up with the default for `touch`.

Is that adequate for you?

LeFrosch:
I understand that you do not want to add these options. Changing the default to 755 for mkdir would also solve my problem. Thanks for the support.

jamestrew:
Thanks, I'll close this over https://github.com/nvim-telescope/telescope-file-browser.nvim/pull/361

PR #367, opened 2024-03-19
Updated Readme

Catacomba:
Added a concrete example of the alternative approach via lua.

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,7 +182,13 @@ vim.api.nvim_set_keymap(
 )
 ```
 
-Alternatively, you can also access the picker as a function via `require "telescope".extensions.file_browser.file_browser` natively in lua.
+Alternatively, you can also access the picker as a function via `require "telescope".extensions.file_browser.file_browser()` natively in lua as follows:
+
+```lua
+vim.keymap.set('n', '<space>fb', function()
+  require 'telescope'.extensions.file_browser.file_browser()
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -182,7 +182,13 @@ vim.api.nvim_set_keymap(
 )
 ```
 
-Alternatively, you can also access the picker as a function via `require "telescope".extensions.file_browser.file_browser` natively in lua.
+Alternatively, you can also access the picker as a function via `require "telescope".extensions.file_browser.file_browser()` natively in lua as follows:
```
jamestrew, line 185: Maybe while we're at it, we can standardize the rest of this keymap section to use `vim.keymap.set` and put it all under one codeblock.

```lua
vim.keymap.set("n", "<space>fb", ":Telescope file_browser<CR>")

-- open file_browser with the path of the current buffer
vim.keymap.set("n", "<space>fb", ":Telescope file_browser path=%:p:h select_buffer=true<CR>")

-- Alternatively, using lua API
vim.keymap.set("n", "<space>fb", function()
	require("telescope").extensions.file_browser.file_browser()
end)
```

jamestrew (changes requested):

README.md
Catacomba: I've standardized keymaps as you requested

jamestrew (approved):

jamestrew:
Thanks!

PR #390, opened 2024-05-20
fix: move has_fd check into a function to reduce startup time

scanhex:
`vim.fn.executable "fd"` takes up to one second, at least on WSL 2 + NixOS setup I'm using. In this PR I'm moving the invocation into a lazy function so that when the plugin is loaded at startup the user wouldn't incur this cost right away.


jamestrew:
Sorry for the delay.
Strange that `vim.fn.executable` takes that long but turning this check lazy and caching is not a bad idea.

Thanks!

PR #404, opened 2024-10-01
feat:  support for creating multiple files using pattern.

Mirsmog:
# Added Multi-File Creation Functionality

   - Now possible to create multiple files at once using the pattern {file1,file2,file3}.ext.
   - Updated action.create and action.create_from_prompt to use create_items for handling multiple files.
   - This allows generating several files in one go.
 [example](https://github.com/user-attachments/assets/20756c25-d18d-4074-8fd8-e8b87139c460)

### Why This Matters:

This change speeds up the workflow by reducing the steps needed to create multiple files. It makes the process quicker and simpler, cutting down repetitive tasks.

I’m not a expert, but I believe this feature will make things easier for others and improve file management efficiency.

### DIFF
--- a/lua/telescope/_extensions/file_browser/actions.lua
+++ b/lua/telescope/_extensions/file_browser/actions.lua
@@ -94,24 +94,37 @@ local create = function(file, finder)
 end
 
 local function create_items(input, finder)
-  if not input then
+  if not input or input == "" then
     return
   end
-  local prefix, middle, suffix = input:match "^(.-){(.-)}(.-)$"
-  local files = {}
 
-  if not (prefix and middle and suffix) then
-    local file = create(input, finder)
-    if file then
-      table.insert(files, file)
+  local function process_pattern(path)
+    local prefix, middle, suffix = path:match "^(.-){(.-)}(.-)$"
+    local results = {}
+
+    if not (prefix and middle and suffix) then
+      return { path }
     end
-    return files
+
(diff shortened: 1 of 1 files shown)


jamestrew:
I like the idea of being able to use bash-style shell expansion to create files/folders.

But re-implementing shell expansions inside of telescope-file-browser however is almost certainly outside the scope of this plugin though. It gets much more complicated than just your implementation of splitting commas separated strings and I'm worried merging this is just going to invite more shell expansion support.
https://www.gnu.org/software/bash/manual/html_node/Brace-Expansion.html

That's not to say it's impossible to support something like this. You could probably do something like creating a custom action to just `os.execute("touch " .. input)`. But even this, I probably wouldn't include into the plugin because it's platform and shell dependent.

Mirsmog:
Thank you for your clarification. I hadn't fully considered that implementing this feature would require supporting the full spectrum of bash logic. Initially, I thought it might be feasible to include just the functionality demonstrated in the example without expanding to support additional bash features. However, you're right in suggesting that this could lead other users to expect complete bash functionality.

Regarding the use of touch via os.execute, I agree that this isn't a solution either, as it would introduce more complexity and dependencies.

Nevertheless, do you think it would be possible to integrate just this specific part of the functionality without adding the rest of the bash expansion logic? This could provide a useful feature while keeping the implementation simple.

Scutua (approved):

Issue #414, opened 2025-05-30
Warning: vim.tbl_flatten is deprecated, use vim.iter(…):flatten():totable() instead.


paulbrittain:
### Description

vim.tbl_flatten is listed as deprecated and can be seen in `checkhealth vim.deprecated`

### Neovim version

```markdown
NVIM v0.12.0
Build type: RelWithDebInfo
LuaJIT 2.1.1744317938
```

### Operating system and version

Linux

### Steps to reproduce

1. Open Neovim
2. Run `require("telescope").extensions.file_browser.file_browser`
3. Observe deprecation warning

or

1. Run `checkhealth vim.deprecated`
2. Observe deprecation warning

### Actual behavior

When running the filebrowser function: `vim.tbl_flatten is deprecated. Run ":checkhealth vim.deprecated" for more information`

Result of `checkhealth vim.deprecated`:
```
- ⚠️ WARNING vim.tbl_flatten is deprecated. Feature will be removed in Nvim 0.13
  - ADVICE:
    - use vim.iter(…):flatten():totable() instead.
    - stack traceback:
        /home/paulbrittain/.local/share/nvim/lazy/telescope-file-browser.nvim/lua/telescope/_extensions/file_browser/fs_stat.lua:83
        /home/paulbrittain/.local/share/nvim/lazy/telescope-file-browser.nvim/lua/telescope/_extensions/file_browser/make_entry.lua:175
        /home/paulbrittain/.local/share/nvim/lazy/telescope.nvim/lua/telescope/pickers/entry_display.lua:155
        /home/paulbrittain/.local/share/nvim/lazy/telescope.nvim/lua/telescope/pickers.lua:1265
        /home/paulbrittain/.local/share/nvim/lazy/telescope.nvim/lua/telescope/entry_manager.lua:129
        /home/paulbrittain/.local/share/nvim/lazy/telescope.nvim/lua/telescope/pickers.lua:1409
        /home/paulbrittain/.local/share/nvim/lazy/telescope.nvim/lua/telescope/pickers.lua:1444
        /home/paulbrittain/.local/share/nvim/lazy/telescope.nvim/lua/telescope/finders/async_static_finder.lua:32
        /home/paulbrittain/.local/share/nvim/lazy/telescope-file-browser.nvim/lua/telescope/_extensions/file_browser/finders.lua:250
[...]

PR #415, opened 2025-05-30
fix: use vim.iter(...):flatten():totable()

paulbrittain:
- fixes: 'Warning: vim.tbl_flatten is deprecated, use vim.iter(…):flatten():totable() instead.'
- replaces usage of vim.tbl_flatten() while keeping backward compatibility for < v0.10.0.

fixes issue: #414


jamestrew:
> ## Pull Request Overview
> This PR addresses the deprecation warning by replacing the usage of vim.tbl_flatten() with vim.iter(...):flatten():totable() for Neovim 0.10+ support while retaining backward compatibility for earlier versions.
> 
> * Replaces the direct use of vim.tbl_flatten() with a conditional that employs the new vim.iter(...):flatten():totable() syntax when available.
> * Maintains support for older Neovim versions by falling back to vim.tbl_flatten().

First time using this. Was just curious.

Thanks

PR #420, opened 2025-08-05
Change folder browser to respect depth

henryxparker:
I think the folder browser should respect the depth option. I'm currently trying to build a project selector in my nvim, and I need the folder browser to respect the depth if my project select is going to be useful. 

Also the closer the two are to feature parity, the more it makes sense for them to be configurations of the same action. The more differences there are, the more confusing it can be when you expect configuration options to be respected. The only real differences in the search should be that one filters out files and defaults to the cwd.

Yes, I could override the search function, but I think this makes more sense from a users perspective

### DIFF
--- a/lua/telescope/_extensions/file_browser/finders.lua
+++ b/lua/telescope/_extensions/file_browser/finders.lua
@@ -73,14 +73,7 @@ local function fd_file_args(opts)
   end
   return args
 end
-
-local function scandir_file_args(opts)
-  local args = {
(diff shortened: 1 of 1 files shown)


henryxparker:
Originally started as [419](https://github.com/nvim-telescope/telescope-file-browser.nvim/pull/419), but I wanted to add it to both supported scans for consistency

henryxparker:
Now that #419  is merged, this is just adding the same thing to the plenary scandir

